Issue #10, opened 2019-03-28
Google Test is fetched two times


riemann89:
Google test dependency is added two times into the project. At the repo level, there is no problem since we have just two references, so no big deal. But when we run `submodule update init --recursive` we download it two times. So it would be nice to remove one of the two occurrences (the one in our dependency folder `depends/googletest`, leaving the one in the libsnark-fork folder in order to diverge as less as possible from the main repo).

AntoineRondelet:
Closing this issue as the corresponding PR has been merged

PR #28, opened 2019-05-01
Removed google duplicated dependency

riemann89:
Fixing #10
In case the dependency has been initialized the user will see a `depends/googletest untracked` folder that has to delete manually.

https://gist.github.com/myusuf3/7f645819ded92bda6677


dtebbs:
Rebased this onto `develop`

dtebbs:
Our submodule import of googletest seems not to be used in any of our cmake files - we use the definitions coming from the libsnark / libfqfft.  This change harmlessly removes our submodule.

PR #44, opened 2019-08-12
Some further fixes for asserts in debug builds

dtebbs:


dtebbs:
One of the asserts shows up what may be a bug in `get_input_bit_size`, after the mimc changes.

src/circuits/joinsplit.tcc
```diff
@@ -94,7 +99,7 @@ class joinsplit_gadget : libsnark::gadget<FieldT> {
                 // The root is represented on a single field element
                 // Each nullifier, and each commitment are in {0,1}^256 and thus take 2 field elements to be represented,
                 // while value_pub_in, and value_pub_out are in {0,1}^64, and thus take a single field element to be represented
-                int nb_inputs = 1 + (2 * (NumInputs + NumOutputs)) + 1 + 1;
+                size_t nb_inputs = (2 * (NumInputs + NumOutputs)) + 1 + 1;
```
AntoineRondelet: Why did you remove the first `+1` @dtebbs ?

The number of primary inputs should be:
`1 (the merkle root =  a single field element) + 2 *(NumInputs + NumOutputs) (Each input nullifier gets packed into 2 field elements, likewise for each output commitment) + 1 (v_pub_in) + 1 (v_pub_out)`, right?

src/circuits/joinsplit.tcc
```diff
@@ -364,9 +369,6 @@ class joinsplit_gadget : libsnark::gadget<FieldT> {
         static size_t get_input_bit_size() {
             size_t acc = 0;
 
-            // Binary length of the Merkle Root (anchor)
-            acc += 256;
```
AntoineRondelet: Right, this one needed to be deleted I think. @rrtoledo can you confirm?
We do not have a `multipacking_gadget` anymore for the root as it is given directly as a field element

AntoineRondelet: Moreover, using `get_input_bit_size` in the assert (https://github.com/clearmatics/zeth/blob/mimc-tree/src/circuits/joinsplit.tcc#L161) is not what we want anymore, since we do not have an entry for the root anymore in the `unpacked_inputs`

src/circuits/joinsplit.tcc
dtebbs: Oh yes.  I forgot that value is also used to setup the protoboard, not just for the packed inputs length.  I'll fix it.

dtebbs: (Sorry about that ... pushed another commit to fix this.  Feel free to squash or replace my commits.)

AntoineRondelet: Thanks

src/circuits/joinsplit.tcc
rrtoledo: Yes this function needs to be updated . Not only do we need to remove the MerkleRoot but we also need to add h_sig +and h_iS.

rrtoledo: > Moreover, using `get_input_bit_size` in the assert (https://github.com/clearmatics/zeth/blob/mimc-tree/src/circuits/joinsplit.tcc#L161) is not what we want anymore, since we do not have an entry for the root anymore in the `unpacked_inputs` (if we want to keep using `get_input_bit_size` in the assert, then we need to subtract the bit_length of the root)

I am not sure what you mean here @AntoineRondelet . `unpacked_inputs` does not include the root, so the assert should still apply once we remove the merkle root from the function.

AntoineRondelet: The function is called `get_input_bit_size` and the merkle root is an input, so there's an inconsistency here between the function name and what's in the body. We need to:
- Either rename the function to "get_unpacked_input_bit_size" and then we can remove the root bit-length from the function (the merkle_root does not appear in the unpacked_input, so it makes sense to remove its mention in the body of the function)
- Or, add the correct bit-length of the merkle root to the `acc` and subtract its length in the assert (the function is called `get_input_bit_size` and the merkle_root is an input so it makes sense to account for it in the body of the function, but then we need to carry out the subtraction of the merkle root length in the assert since we assert on the unpacked_inputs)

PR #45, opened 2019-08-12
Fix unit tests

dtebbs:


dtebbs:
May be specific to mac compilers, but the merkle tree tests was not being run because the main function was in the anonymous namespace (showed up as a compile warning that `main` was unused).  When I moved it out of the namespace, it started failing:

```
Not valid!
Expected 15111851447014879833050233394183206021293104970044755574134456851342505158717
Actual 7121700468981037559893852455893095765125417767594185027454590493596569372187
[test_merkle_path_authenticator_depth1] Expected (True), Obtained result: 0
/Users/duncantebbs/dev/zeth/src/test/merkle_tree_test.cpp:215: Failure
Value of: res
  Actual: false
Expected: true
[  FAILED  ] MainTests.TestMerkleTreeField (4 ms)
```

rrtoledo:
Thks for noticing!

There was an issue with the namespace.

As for the fact that the test doesn't pass, it is because of old values (as you can see there is a variable iv which is not used anymore).

AntoineRondelet:
Are you still planning to work on this branch @dtebbs (ie: is this PR still WIP)?

dtebbs:
@AntoineRondelet I was not planning to do anything more here.
My intention was just to leave this open for the test fix, which @rrtoledo seems to have pushed.
So as far as I'm concerned, I think it's ready.

Issue #46, opened 2019-08-12
Dependencies in cmake files


dtebbs:
Source file and dependency specifications seem to suffer from a few problems:
- source files are re-specified for every executable (utility and unit test) which inflates build times
- some include files and .tcc files are not specified as belonging to any target which means that incremental builds (using Makefiles) are often incomplete.

AntoineRondelet:
Closed as the corresponding PR has been merged.
Thanks for that @dtebbs

PR #47, opened 2019-08-12
Tidy up some cmake files

dtebbs:
Addresses #46


AntoineRondelet:
Let's look at #50 first and merge this one after the resolution of the problem with `test_prover` and `libboost`.

dtebbs:
> Let's look at #50 first and merge this one after the resolution of the problem with test_prover and libboost.

Yes.  Sorry about that.  For some reason, I didn't need to link boost_system on my machine (mac difference, maybe), so I didn't include it.
I pushed a slightly different way of specifying the boost_system dependency, but feel free to force-push or replace with your PR if you prefer.

AntoineRondelet:
> I pushed a slightly different way of specifying the boost_system dependency, but feel free to force-push or replace with your PR if you prefer.

Have you pushed on this branch? I can't see any new commits to link `boost_system`

dtebbs:
Sorry, the push didn't work.  Should be there now.

Issue #49, opened 2019-08-13
Switch sha256 to Blake2s


AntoineRondelet:
Blake2s (see: https://blake2.net/) looks more efficient - constraint-wise - than sha256. Hence, we should consider switching to this hash function when possible to diminish the number of multiplicative gates in the arithmetic circuit.

Related discussion: https://github.com/zcash/zcash/issues/2630

AntoineRondelet:
Closing as the corresponding PR was merged

PR #63, opened 2019-08-23
Blake2s update

rrtoledo:


src/circuits/blake2s/blake2s_comp.tcc
```diff
@@ -0,0 +1,155 @@
...
+}
+
+template<typename FieldT>
+size_t BLAKE2s_256_comp<FieldT>::get_block_len() {
+    return BLAKE2s_block_size;
+}
+
+template<typename FieldT>
+size_t BLAKE2s_256_comp<FieldT>::expected_constraints(const bool ensure_output_bitness)
+{
```
rrtoledo: fixed

AntoineRondelet:
This PR tackles issue https://github.com/clearmatics/zeth/issues/49

pyClient/zethGRPC.py
```diff
@@ -175,18 +175,18 @@ def hexadecimalDigestToBinaryString(digest):
...
 
-# Returns nf = sha256(1110 || [a_sk]_252 || rho)
+# Returns nf = blake2s(1110 || [a_sk]_252 || rho)
 def computeNullifier(zethNote, spendingAuthAsk):
     binaryAsk = hexadecimalDigestToBinaryString(spendingAuthAsk)
     first252Ask = binaryAsk[:252]
     leftLegBin = "1110" + first252Ask
     leftLegHex = "{0:0>4X}".format(int(leftLegBin, 2))
-    nullifier = hashlib.sha256(
+    nullifier = blake2s(
```
AntoineRondelet: The nullifier should be computed by a call to the compression function of blake2s, not the full hash, right? Don't you use the full hash here?

src/circuits/blake2s/blake2s_comp.hpp
```diff
@@ -0,0 +1,74 @@
...
+    std::array<std::array<libsnark::pb_variable_array<FieldT>, 16>, rounds>
+        v_temp;
+    std::array<libsnark::pb_variable_array<FieldT>, 8> output_bytes;
+    libsnark::block_variable<FieldT> input_block;
+    libsnark::digest_variable<FieldT> output;
+
+    std::array<std::vector<g_primitive<FieldT>>, 10> g_arrays;
+    std::vector<xor_constant_gadget<FieldT>> xor_vector;
+
+    // TODO: Remove ZERO and pass it in the constructor
```
AntoineRondelet: Can you pass the ZERO to the constructor?

src/circuits/blake2s/blake2s_setup.tcc
```diff
@@ -0,0 +1,303 @@
+#ifndef __ZETH_BLAKE2S_SETUP_TCC__
+#define __ZETH_BLAKE2S_SETUP_TCC__
+
+namespace libzeth
+{
+
+template<typename FieldT> void BLAKE2s_256_comp<FieldT>::setup_constants()
```
AntoineRondelet: Add reference paper to know where the constants come from

src/circuits/blake2s/blake2s_setup.tcc
```diff
@@ -0,0 +1,303 @@
+#ifndef __ZETH_BLAKE2S_SETUP_TCC__
```
AntoineRondelet: There does not seem to be the corresponding `hpp` file.
Please add the `.hpp` file to be consistent with the rest of the code, include this tcc file after closing the `libzeth` namespace and add reference paper in the comments at the beginning of the file.

src/circuits/circuits-util.tcc
```diff
@@ -102,6 +102,70 @@ libsnark::pb_variable_array<FieldT> from_bits(
     return acc;
 };
 
+// Sum 2 binary strings, discard last carry
+template<typename FieldT, size_t BitLen>
+std::array<FieldT, BitLen> binaryFieldAdditionNoCarry(
```
AntoineRondelet: Switch to `snake_case` for the naming

src/circuits/circuits-util.tcc
```diff
@@ -102,6 +102,70 @@ libsnark::pb_variable_array<FieldT> from_bits(
...
+        carry = A[i] * B[i] + A[i] * carry + B[i] * carry -
+                FieldT(2) * A[i] * B[i] * carry;
+    }
+
+    return sum;
+}
+
+// XOR 2 binary strings
+template<typename FieldT, size_t BitLen>
+std::array<FieldT, BitLen> binaryFieldXOR(
```
AntoineRondelet: Switch to `snake_case` for the naming

src/circuits/circuits-util.tcc
```diff
@@ -102,6 +102,70 @@ libsnark::pb_variable_array<FieldT> from_bits(
...
+    std::array<FieldT, BitLen> xor_array;
+    xor_array.fill(FieldT("0"));
+    for (size_t i = 0; i < BitLen; i++) {
+        xor_array[i] = A[i] + B[i] - FieldT(2) * A[i] * B[i]; // c is carry
+    }
+
+    return xor_array;
+}
+
+template<typename FieldT> std::vector<FieldT> convertToBinary(size_t n)
```
AntoineRondelet: Switch to `snake_case` for the naming

src/circuits/commitments/commitments.hpp
```diff
@@ -47,8 +46,8 @@ libsnark::pb_variable_array<FieldT> getRightSideCMCOMM(
 // the value of the commitment_k without needing 2 distinct gadgets for this.
 //
 // See Zerocash extended paper, page 22
-// The commitment k is computed as k = sha256(r || [sha256(a_pk || rho)]_128)
-// where we define the left part: inner_k = sha256(a_pk || rho)
+// The commitment k is computed as k = blake2s(r || [blake2s(a_pk || rho)]_128)
+// where we define the left part: inner_k = blake2s(a_pk || rho)
```
AntoineRondelet: `blake2sCompress` (change in other relevant places)

src/circuits/joinsplit.tcc
```diff
@@ -619,11 +618,17 @@ public:
         // < FieldT::capacity()
         nb_elements += 1;
 
-        // We allocate 2 field elements to pack the value of h_sig
-        nb_elements += 2;
+        // h_sig is represented by 2 field element (if we consider a digest_len
```
AntoineRondelet: Why is this showing up in this PR? This is related to the transaction malleability fix which has already been merged to develop

src/circuits/prfs/prfs.hpp
```diff
@@ -78,8 +76,8 @@ class PRF_addr_a_pk_gadget : public PRF_gadget<FieldT, HashT>
 };
 
 // PRF to generate the nullifier
-// nf = sha256("1110" || [a_sk]_252 || rho): See ZCash protocol specification
-// paper, page 57
+// nf = blake2s("1110" || [a_sk]_252 || rho): See ZCash protocol
```
AntoineRondelet: We use the compression function here, it should be `blake2sCompress`

src/circuits/simple_gadgets.hpp
```diff
@@ -0,0 +1,110 @@
+#ifndef __ZETH_SIMPLE_GADGETS_HPP__
```
AntoineRondelet: `simple_gadgets.hpp` is quite general and does not very well inform on the content of the document. Please rename accordingly (maybe something like `xor_gadgets.hpp` or `blake2s_(components|operations)`?)

src/test/mimc_test.cpp
```diff
@@ -13,7 +15,147 @@ typedef libff::Fr<ppT> FieldT;
 namespace
 {
 
-TEST(TestRound, TestTrue)
+// Testing that (15212  + 98645 + 216319)**7 =
```
AntoineRondelet: Why is this MiMC file showing up here?

src/test/prfs_test.cpp
```diff
@@ -359,15 +359,15 @@ TEST(TestPRFs, TestPRFPKGadget)
...
     // h_i should equal:
     // 7ea1525fdbf9462c5144796937e1f80b9dad42369f7d4987c436b2f79257f9ac h_i =
-    // sha256( 0i00 || [a_sk]_252 || h_sig) a_sk:
+    // blake2s( 0i00 || [a_sk]_252 || h_sig) a_sk:
     // 0x0F000000000000FF00000000000000FF00000000000000FF00000000000000FF h_sig:
     // 0x0F000000000000FF00000000000000FF00000000000000FF00000000000000FF Note:
-    // This test vector has been generated by using the solidity sha256 function
-    // (we want to make sure that we generate the same digests both on-chain and
-    // off-chain))
+    // This test vector has been generated by using the solidity blake2s
```
AntoineRondelet: Have you generated this test vector by using a `blake2s` solidity implementation?

src/test/sha256_test.cpp
```diff
@@ -0,0 +1,207 @@
+#include "gtest/gtest.h"
```
AntoineRondelet: Why is this test showing up in this PR?

src/test/simple_gadget_test.cpp
```diff
@@ -0,0 +1,270 @@
...
+#include "snarks_alias.hpp"
+
+#include "gtest/gtest.h"
+#include <libff/common/default_types/ec_pp.hpp>
+
+// Access the `from_bits` function and other utils
+#include "circuits/circuits-util.hpp"
+#include "util.hpp"
+
+using namespace libsnark;
```
AntoineRondelet: Remove this `using` directive. Let's avoid to use such directives (Moreover, the namespace prefix seems to be used in the tests body). `using namespace libzeth` is ok as we test functionalities of this library here.

src/types/bits.tcc
```diff
@@ -51,6 +51,38 @@ std::array<bool, BitLen> binaryAddition(
     return sum;
 }
 
+// Sum 2 binary strings, discard last carry
+template<size_t BitLen>
+std::array<bool, BitLen> binaryAdditionNoCarry(
```
AntoineRondelet: Use `snake_case` for naming

src/types/bits.tcc
```diff
@@ -51,6 +51,38 @@ std::array<bool, BitLen> binaryAddition(
...
+        sum[i] = ((A[i] ^ B[i]) ^ carry); // c is carry
+        carry = ((A[i] & B[i]) | (A[i] & carry)) | (B[i] & carry);
+    }
+
+    return sum;
+}
+
+// XOR 2 binary strings
+template<size_t BitLen>
+std::array<bool, BitLen> binaryXOR(
```
AntoineRondelet: Use `snake_case` for naming

src/types/bits.tcc
AntoineRondelet: Could you also rename the `binaryAddition` too? That way it'll all be consistent.

pyClient/zethGRPC.py
rrtoledo: > Don't you use the full hash here?
You are correct. 

Using the full hash on 2 inputs of size _bloc\_size_ is equivalent to using the compression function.

src/circuits/blake2s/blake2s_comp.hpp
rrtoledo: I started doing this in the fix-todos branch (c.f. sha256, joinsplit circuits...) . I intended to do this on that branch once the blake2s PR is finished.

src/test/prfs_test.cpp
rrtoledo: No, I forgot to change the word "solidity" by hashlib (it should be the case in other files, I'll check and correct that).

src/test/sha256_test.cpp
rrtoledo: When doing tests on blake2s I saw that the sha256 were not correctly implemented or called. I took the opportunity to fix that (even if we now are not using sha256 anymore (irony)).

src/test/prfs_test.cpp
AntoineRondelet: Thanks

src/circuits/blake2s/blake2s_comp.hpp
AntoineRondelet: As you prefer as long as we do not forget :)

src/types/bits.tcc
rrtoledo: There is already a binaryAddition function which throws an error if there is a carry at the last bit. I was not sure whether we wanted to distinguish the two uses so I created this function.

What would you think of adding a boolean to binaryAddition so that it does not throw an exception when needed ?

src/circuits/blake2s/blake2s_setup.tcc
rrtoledo: The functions are alreadu defined in blake2s_comp.hpp . I separated these functions (setup & computation) for clarity (since the file would have been very long). Do you still want an .hpp ?

I could also rename this file as blake2s_comp.cpp as it was done for /zeth/src/types/bits.cpp .

src/test/sha256_test.cpp
AntoineRondelet: > When doing tests on blake2s I saw that the sha256 were not correctly implemented or called

What was wrong with the sha256 gadget? Do you remember the changes you made? The diffs only show the full file because of the rebase I assume

src/types/bits.tcc
AntoineRondelet: Yes you can add a flag as arg to the function: https://github.com/clearmatics/zeth/blob/master/src/types/bits.tcc#L28
To determine whether we mod by `2^BitLen` or not. Please set the default value of the argument to be "false" to avoid bad surprises. You'll need to add the flag as part of the conditional to throw the error also.
(and please add a comment above the function explaining the role of the `mod` flag)

src/circuits/blake2s/blake2s_setup.tcc
AntoineRondelet: > I could also rename this file as blake2s_comp.cpp as it was done for /zeth/src/types/bits.cpp

Let's keep the `.tcc` extension as this is a templatized file.

> The functions are alreadu defined in blake2s_comp.hpp . I separated these functions (setup & computation) for clarity (since the file would have been very long). Do you still want an .hpp

Then maybe rename the file `blake2s_comp_setup.tcc` so we know it's part of the `comp` group of files

src/circuits/blake2s/blake2s_comp.tcc
```diff
@@ -0,0 +1,178 @@
...
+    //                                   &annotation_prefix) )
+
+    // Push the block variable in local to be swapped
+    std::vector<FieldT> padded_input;
+    for (size_t i = 0; i < input_size; i++) {
+        padded_input.push_back(this->pb.val(input_block.bits[i]));
+    }
+
+    // [SANITY CHECK] Pad if necessary
+    for (size_t i = 0; i < 512 - input_size; i++) {
```
AntoineRondelet: 512 is the block_size. Please replace by the constant name associated to it (ie: `BLAKE2s_block_size`)

src/circuits/blake2s/blake2s_setup.tcc
AntoineRondelet: ie: https://blake2.net/blake2.pdf appendix A.2

src/test/sha256_test.cpp
rrtoledo: The gadget was fine. It was the test file which had wrong values (and if I remember well some problems with pointers).

AntoineRondelet: What kind of problems with pointers? 
So the test values were wrong, but the tests were passing?

src/test/mimc_test.cpp
AntoineRondelet: Any news on that @rrtoledo?

src/circuits/joinsplit.tcc
AntoineRondelet: Any news on that @rrtoledo?

src/circuits/blake2s/blake2s_comp.hpp
rrtoledo: Done in fix-todo PR

src/test/sha256_test.cpp
rrtoledo: I have just checked and I do not  see any modification between this version and the one in develop. I do not really understand.

What I recall is that the test was not called and I had to make a few modifications. 

There was a problem with the values and I thought that was due to their initialization, the pointers or the init_public_param. I cannot recall.

AntoineRondelet: We'll need to investigate that because we can't merge it without understanding why the file has been changed and what modifs were made (let's follow a "1 PR-1 thing" approach so that we know exactly what is merged)

src/test/mimc_test.cpp
rrtoledo: I merged round, mimc_mp_test and mimc_test into one file (mimc_test) to clean up as we have more and more tests.

src/circuits/joinsplit.tcc
rrtoledo: I wanted the comments to be consistent, the parenthesis was missing.

src/circuits/simple_gadgets.hpp
rrtoledo: I wanted to keep a name independent of blake in case we were to reuse it. xor_gadgets would be great if not for the "double_bit32_sum_eq_gadget".

What about something like binary_operation_gadget or herlper_gadgets ?

src/test/sha256_test.cpp
rrtoledo: Should I overwrite this doc with develop's version then?

src/test/mimc_test.cpp
AntoineRondelet: Please do this in a separate PR as this is not related to blake2s

src/test/sha256_test.cpp
AntoineRondelet: Let's find out what changes you've done (you can compare the files on this branch and on develop to figure this out I guess) and move this into a separate PR as this is not related to blake2s

AntoineRondelet:
@rrtoledo I did some cleanup and documentation of the code here: https://github.com/clearmatics/zeth/pull/81

src/circuits/simple_gadgets.hpp
AntoineRondelet: ` binary_operation_gadgets` sounds more indicative

src/test/sha256_test.cpp
rrtoledo: This shows up because I wanted to rename test_sha256 into sha256_test as all the tests are written in that fashion ("xxx_test").

The current test_sha256 compilation fails, because there is no annotation given during the variable allocation and the interface of HashT must have somehow changed  since we have "libsnark::SHA256_block_size" as an input of the constructor in the tests (which is wrong).

The modifications I made includes changing the name (it created a new file and should have deleted test_sha256 but I can still see it), removing this sha256_block_size, adding the annotations and I am not sure why replacing pointer variables by variables.

src/circuits/simple_gadgets.hpp
rrtoledo: ok !

src/test/sha256_test.cpp
rrtoledo: I'll open a new PR to enforce these changes (not the pointers part) as said previously.

rrtoledo: Issue created, and changes reverted.

AntoineRondelet: Please make sure to point to the issue for the record. ie: https://github.com/clearmatics/zeth/issues/83

src/test/mimc_test.cpp
AntoineRondelet: Related issue opened https://github.com/clearmatics/zeth/issues/84, will be fixed in a sperate PR

src/test/commitments_test.cpp
```diff
@@ -481,19 +479,18 @@ TEST(TestCOMMs, TestCOMMCMGadget)
...
     //
     // outer_k:
     // 0xAF000000000000FF00000000000000FF00000000000000FF00000000000000FF '0^192
     // || value_v':
     // 0x0000000000000000000000000000000000000000000000002F0000000000000F
     //
-    // Note: This test vector has been generated by using the solidity sha256
-    // function (we want to make sure that we generate the same digests both
-    // on-chain and off-chain) Solidity version used: v0.5.0
+    // Note: This test vector has been generated by using the hashlib sha256
```
AntoineRondelet: Why is sha256 showing up here?

src/test/commitments_test.cpp
```diff
@@ -382,16 +381,15 @@ TEST(TestCOMMs, TestCOMMOuterKGadget)
...
     //
     // Hex of the 384 bits of the trap_r:
     // "0x0F000000000000FF00000000000000FF00000000000000FF00000000000000FF00000000000000FF00000000000000FF"
     // Hex of the first 128bits of the inner_k:
     // "0xEF000000000000FF00000000000000FF"
     //
-    // Note: This test vector has been generated by using the solidity sha256
-    // function (we want to make sure that we generate the same digests both
-    // on-chain and off-chain) hash generated with solidity v0.5.0:
+    // Note: This test vector has been generated by using the hashlib sha256
```
AntoineRondelet: Shouldn't it be blake2s instead?

src/test/commitments_test.cpp
rrtoledo: Just updated with blake2s. I must have missed these notes.

src/test/commitments_test.cpp
rrtoledo: Likewise (updated).

src/circuits/circuits-util.tcc
```diff
@@ -102,6 +102,70 @@ libsnark::pb_variable_array<FieldT> from_bits(
     return acc;
 };
 
+// Sum 2 binary strings, discard last carry
+template<typename FieldT, size_t BitLen>
+std::array<FieldT, BitLen> binary_field_addition_no_carry(
```
rrtoledo: @AntoineRondelet We have similar functions (xor, addition...) working on _bits_ and not_fields_ in types/bits.tcc.
Would you agree to merge circuits-util.* into bits.* ?

AntoineRondelet: > Would you agree to merge circuits-util.* into bits.* ?

I'm not sure I understand what you mean @rrtoledo. The circuits-util.* files aim to provide util functions for the circuit and manipulate bits represented by FieldT as shown on the line you commented. The `bits.*` files on the other hand, are used to introduce new types and functions that are useful to manipulate them.

AntoineRondelet: The `circuit-utils` aim to be called inside the circuit while the others can be used outside

rrtoledo: Fair enough.

I did not see this distinction between the files, only that they both process bits (encoded as bool or FieldT).

src/test/blake2s_test.cpp
```diff
@@ -27,37 +27,60 @@ TEST(TestG, TestTrue)
...
+        {
+            0, 1, 0, 1, 0, 0, 0, 1, // 51
+            0, 0, 0, 0, 1, 1, 1, 0, // 0E
+            0, 1, 0, 1, 0, 0, 1, 0, // 52
+            0, 1, 1, 1, 0, 1, 0, 0  // 74
+        },
         ZERO);
 
-    // 0110 1100  0110 1100  0110 0101  0110 1000
+    // "lleh"
```
AntoineRondelet: Where does this come from @rrtoledo? (same q for "ow o" below)

rrtoledo: This represents the first call of G_primitive in blake2s compression function when hashing "hello world".
If you remember blake2s compression function will split the input block into 32bit ("hell", "o wo", "rld") words in little endian ("lleh", "ow o", "dlr") and then process then 2 at a time ("lleh" and "ow o"). 

I will add these details here as the mention of "lleh" and "ow o" alone is confusing.

rrtoledo: `// The test correponds to the first call of the g_primitive of blake2s(b"hello world")
// As blake2s first format the input blocks in 32bit words in little endian,
// the inputs of the firs G_Primitive are "lleh" and "ow o"
// ("hello world" -> "hell" "o wo" "rld" plus padding-> "lleh" "ow o" "dlr" plus padding)
TEST(TestG, TestTrue)`

src/test/prover_test.cpp
```diff
@@ -704,25 +698,24 @@ bool TestInvalidJS2In2(
...
-    bits256 rho_out_bits256 = get_bits256_from_vector(
-        hexadecimal_digest_to_binary_vector("1111000000000000000000000000000000"
-                                            "000000000000000000000000009777"));
+    bits256 rho_out_bits256;
     bits384 trap_r_out_bits384 =
         get_bits384_from_vector(hexadecimal_str_to_binary_vector(
             "11000000000000990000000000000099000000000000007700000000000000FF00"
             "000000000000FF0000000000000777"));
+
+    // // 0x8530000A00000000 = 9.597170848876199937 eTH
```
rrtoledo: The extra // are removed

rrtoledo: (and eTH capitalized)

src/test/blake2s_test.cpp
AntoineRondelet: > I will add these details here as the mention of "lleh" and "ow o" alone is confusing.

yes, thanks. Perfect, the comment you dropped right after clarifies this. we should take the habit to add comments where necessary so that the code base never becomes a "mystery" (and this saves time for when we come back these files in the future)

src/circuits/binary_operation_gadgets.hpp
```diff
@@ -0,0 +1,112 @@
...
+        const std::string &annotation_prefix = "double_bit32_sum_eq_gadget");
+
+    void generate_r1cs_constraints(bool enforce_boolean = true);
+    void generate_r1cs_witness();
+};
+
+} // namespace libzeth
+#include "binary_operation_gadgets.tcc"
+
+#endif // __ZZETH_SIMPLE_GADGETS_HPP__
```
AntoineRondelet: `ZZETH` -> `ZETH`

src/test/blake2s_test.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+    // Second word in little endian "ow o"
+    libsnark::pb_variable_array<FieldT> y = from_bits(
+        {
+            0, 1, 1, 0, 1, 1, 1, 1, // 6F
+            0, 1, 1, 1, 0, 1, 1, 1, // 77
+            0, 0, 1, 0, 0, 0, 0, 0, // 20
+            0, 1, 1, 0, 1, 1, 1, 1  // 6F
+        },
+        ZERO);
```
AntoineRondelet: So, you do not pass the final block "dlr" to the hash function to retrieve the hash of "hello world"?

rrtoledo: This test tests g_primitive which only takes as inputs some state (a,b,c,d here) and _two_ words (x and y here). 

I made the choice to choose as test vector the first g_primitive of the first round of the compression of block input "hello world". In that case g_primitive takes as input only the first two words ("lleh" and "ow o".

If I want to test the final block "dlr", it means I am testing  g_primitive another time which I thought was not necessary.

If you want to add this test, I need to pad "dlr", define y as some padding, update the state (a,b,c,d) as the output of the first g_primitive gadget and compute the output.

src/CMakeLists.txt
```diff
@@ -137,14 +137,17 @@ endfunction(zeth_test)
 zeth_test(test_simple test/simple_test.cpp TRUE)
 zeth_test(test_addition test/packed_addition_test.cpp TRUE)
 zeth_test(test_hex_to_field test/hex_to_field_test.cpp TRUE)
-zeth_test(test_note test/note_test.cpp TRUE)
-zeth_test(test_prfs test/prfs_test.cpp TRUE)
-zeth_test(test_commitments test/commitments_test.cpp TRUE)
+zeth_test(test_binary_operation test/binary_operation_test.cpp TRUE)
+# zeth_test(test_sha256 test/test_sha256.cpp TRUE)
```
AntoineRondelet: Why did you comment this one out @rrtoledo ?

rrtoledo: I added it because it was missing. 

You asked me to remove it as we do not use this circuit any longer.

AntoineRondelet: Not sure I understand, the test for sha is not ran here: https://github.com/clearmatics/zeth/blob/develop/src/CMakeLists.txt#L137-L147, so why do you add it and comment it?

src/test/blake2s_test.cpp
```diff
@@ -0,0 +1,234 @@
...
+        pb, {pb_va_input}, "blake2s_block_input");
+
+    libsnark::digest_variable<FieldT> output(pb, BLAKE2s_digest_size, "output");
+
+    BLAKE2s_256_comp<FieldT> blake2s_comp_gadget(pb, input, output);
+    blake2s_comp_gadget.generate_r1cs_constraints();
+    blake2s_comp_gadget.generate_r1cs_witness();
+
+    // blake2s(b"hello world")
+    libsnark::pb_variable_array<FieldT> expected = from_bits(
```
AntoineRondelet: What did you use to generate the test vectors @rrtoledo?

rrtoledo: I used hashlib for `TestBlake2sComp`.
As for `TestG`, I used a python script (and added prints) referred by blake2.net (https://github.com/buggywhip/blake2_py) .

AntoineRondelet: Can you add a line to specify this in the code so it's clear where these values come from?

rrtoledo: Done

AntoineRondelet: Thanks

PR #64, opened 2019-08-27
Code format manual fixes

AntoineRondelet:
Moved a bunch of comments above lines to avoid breaks where it is not needed, and fixed a few inconsistencies


dtebbs:
That's great.  Thanks.

Issue #60, opened 2019-08-21
Set up Travis CI


AntoineRondelet:
Let's set up travis CI to ease reviews and make sure we keep high standards on the code being merged in the repository.
A minimal configuration of Travis would:
- Run the tests
- Check that the code follows the style guide (TBD, see https://github.com/clearmatics/zeth/issues/57)

AntoineRondelet:
Closing this issue as the corresponding PR has been merged

PR #65, opened 2019-08-28
Enable travis to check the build, tests and formatting

dtebbs:
This PR sets up travis to check code formatting, check the build (including compile warnings) and run some fast-running tests on mac and linux.

Addresses issue #60.

All builds execute with "-Werror" to detect compiler warnings.
The linux build pulls the docker container and runs build and tests inside that.  It runs the code checks natively (relying on git to notice any changes).  The mac build uses the default aple compilers.  In the future it would be nice to use the llvm compilers on mac.

### DIFF
--- a/scripts/ci
+++ b/scripts/ci
@@ -36,7 +36,7 @@ function build() {
         cxx_flags="${cxx_flags} -Wno-deprecated-declarations"
     fi
 
-    cmake_flags="-DFAST_TESTS_ONLY=ON -DCMAKE_BUILD_TYPE=${build_type}"
+    cmake_flags="-DCMAKE_BUILD_TYPE=${build_type}"
 
     . setup_env.sh
     mkdir -p build
--- a/scripts/ci
+++ b/scripts/ci
@@ -7,6 +7,17 @@ echo "running against commit: "`git log --oneline --no-decorate -n 1`
 set -x
 set -e
 
+function format_check() {
+    scripts/format
+    git diff --no-ext-diff | head -n 20 > format_errors
+    num_lines=`cat format_errors | wc -l`
+    if [ "0" != "${num_lines}" ] ; then
+        echo CODE FORMATTING ERRORS:
+        cat format_errors
+        exit 1
+    fi
+}
+
 function build() {
     dir_name=$1
     build_type=$2
@@ -49,6 +60,10 @@ function build_debug() {
 
 # The CI_EXECTUTE_IN_DOCKER variable determines whether we should
 # re-execute in the docker container.
+
+if [ "1" == "${CI_CHECK_FORMAT}" ] ; then
+    format_check
+fi
 
 if [ "1" == "${CI_USE_DOCKER}" ] ; then
     docker pull clearmatics/zeth-base:latest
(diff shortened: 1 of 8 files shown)


dtebbs:
If there are other configs we should be testing then let me know.

scripts/ci
```diff
@@ -0,0 +1,75 @@
...
+    num_lines=`cat format_errors | wc -l`
+    if [ "0" != "${num_lines}" ] ; then
+        echo CODE FORMATTING ERRORS:
+        cat format_errors
+        exit 1
+    fi
+}
+
+function build() {
+
```
AntoineRondelet: Extra line

scripts/ci
```diff
@@ -0,0 +1,75 @@
...
+        export PATH="/usr/local/opt/llvm/bin:/usr/local/bin:${PATH}"
+        export PKG_CONFIG_PATH="/usr/local/opt/openssl/lib/pkgconfig"
+        export LIBRARY_PATH="/usr/local/opt/openssl/lib"
+        export LDFLAGS="-L/usr/local/opt/llvm/lib -L-L/usr/local/lib"
+        export CPPFLAGS="-I/usr/local/opt/llvm/include -I/usr/local/include"
+
+        cxx_flags="${cxx_flags} -Wno-deprecated-declarations"
+    fi
+
+    cmake_flags="-DFAST_TESTS_ONLY=ON -DCMAKE_BUILD_TYPE=${build_type}"
```
AntoineRondelet: Why skip the prover tests? It's a slow one but this is the one that tests the entire circuit.

src/test/simple_test.tcc
```diff
@@ -32,19 +32,19 @@ template<typename FieldT> void simple_circuit(libsnark::protoboard<FieldT> &pb)
     pb.set_input_sizes(1);
 
     //   g1
-    //  /  \
+    //  /  \ .
```
AntoineRondelet: Why did you add `.`?

scripts/ci
AntoineRondelet: We want to make sure all tests pass before merging

scripts/ci
dtebbs: Removed extra line

scripts/ci
dtebbs: I've enabled the prover_test as well.
I originally disabled it for these per-commit builds to get faster turnaround time on results, and wasn't sure if travis had some timeout.  But let's see how it goes.  Definitely better if we can run it on every check-in.

src/test/simple_test.tcc
dtebbs: Yeah, this is quite annoying.  One of the compilers was giving warnings about multi-line comments if there was a `\` at the end of a comment.  I could either try to disable that warning, or adjust the form of the comments if you prefer.

dtebbs: Afterall, I changed the comments to not use `\` at the end.

scripts/ci
dtebbs: prover_test is enabled and running now (builds take approx 22 mins).

AntoineRondelet: Yes, let's see how it goes. Thanks!

PR #73, opened 2019-09-10
[MPC] Phase2 MPC

dtebbs:
Functions and cli utils to perform the phase2 MPC.
Some things that are still WIP / need discussion:
- sources of randomness
- read/write compressed points
- hashing to G2
- robust file operations

(depends on #68)


src/mpc/mpc_create_keypair.cpp
```diff
@@ -130,12 +131,13 @@ class mpc_create_keypair : public subcommand
...
 
         libff::enter_block("Load phase2 data");
-        if (!libff::inhibit_profiling_info) {
-            libff::print_indent();
-            std::cout << phase2_file << "\n";
-        }
-        srs_mpc_layer_C2<ppT> phase2 =
-            read_from_file<srs_mpc_layer_C2<ppT>>(phase2_file);
+        libff::print_indent();
+        std::cout << phase2_file << std::endl;
```
AntoineRondelet: I'd rename `phase2_file` as `phase2_challenge_file` since this file really is the challenge, right?

src/mpc/mpc_dummy_phase2.cpp
```diff
@@ -13,8 +14,8 @@ namespace
 //     mpc dummy_phase2 [<option>] <linear_combination_file>
 //
 // Options:
-//     -h,--help           This message
-//     --out <file>        Write phase2 to <file> (mpc-phase2.bin)
+//     -h,--help       This message
```
AntoineRondelet: `This message`? Why not just say `-h,--help       Displays help message` or something like that?

src/mpc/mpc_dummy_phase2.cpp
```diff
@@ -78,20 +79,22 @@ class mpc_dummy_phase2 : public subcommand
...
-                pb.get_constraint_system();
-            return cs.num_inputs();
+            return pb.num_inputs();
         }();
+        libff::print_indent();
+        std::cout << std::to_string(num_inputs) << std::endl;
+        libff::leave_block("computing num_inputs");
 
-        // Generate the artifical delta
+        // Generate the artificial delta
```
AntoineRondelet: `Generate the artificial delta` -> `Generate a random delta locally to run the dummy phase2`?

depends/libsodium/crypto_stream/chacha20/ref/chacha20_ref.c
```diff
@@ -0,0 +1,312 @@
+
+/*
```
AntoineRondelet: Why not use libsodium as a submodule?

AntoineRondelet: Did you make any changes to the files of libsodium? If so, which ones?

scripts/ci
```diff
@@ -33,13 +34,17 @@ function build() {
...
-        export CPPFLAGS="-I/usr/local/opt/llvm/include -I/usr/local/include"
+        export LDFLAGS="-L/usr/local/lib"
+        export CPPFLAGS="-I/usr/local/include"
 
+        cxx_flags="${cxx_flags} -I/usr/local/opt/openssl/include"
         cxx_flags="${cxx_flags} -Wno-deprecated-declarations"
     fi
 
     cmake_flags="-DCMAKE_BUILD_TYPE=${build_type}"
+    if ! [ "1" == "${full}" ] ; then
```
AntoineRondelet: `"1" == "${full}"` let's switch it to `"${full}" == "1"` as it reads more naturally -> "if full is true/1" instead of "if 1/true is full"

scripts/ci
```diff
@@ -55,11 +60,21 @@ function build() {
...
+full_build=$2
+if [ "" == "${full_build}" ] ; then
+    if [ "false" == "${TRAVIS_PULL_REQUEST}" ] || [ "" == "${TRAVIS_PULL_REQUEST}" ] ; then
+        full_build=0
+    else
+        full_build=1
+    fi
+fi
+
+if [ "1" == "${CI_CHECK_FORMAT}" ] ; then
```
AntoineRondelet: Likewise, let's switch the condition to `"${CI_CHECK_FORMAT}" == "1"`

scripts/ci
```diff
@@ -55,11 +60,21 @@ function build() {
...
 function ci() {
-    build $1
+    build $1 $2
     command_tests
 }
 
-if [ "${CI_CHECK_FORMAT}" == "1" ] ; then
+build_type=$1
+full_build=$2
+if [ "" == "${full_build}" ] ; then
```
AntoineRondelet: `"" == "${full_build}"` -> `"${full_build}" == ""`

AntoineRondelet: this applies to the conditionals below too

scripts/test_mpc
```diff
@@ -4,42 +4,112 @@ set -x
...
+       ${challenge_0_file} ${transcript_file} ${challenge_3_file}
+${MPC} phase2-verify-transcript \
+       --digest ${response_digest_2_file} \
+       ${challenge_0_file} ${transcript_file} ${challenge_3_file}
+${MPC} phase2-verify-transcript \
+       --digest ${response_digest_3_file} \
+       ${challenge_0_file} ${transcript_file} ${challenge_3_file}
+
+# Invalid contribution check
+echo 00000000 >> ${invalid_response_digest_file}
```
AntoineRondelet: Why writing `00000000` in the file?

scripts/test_mpc
```diff
@@ -4,42 +4,112 @@ set -x
...
+
+# Clean up files
+rm ${pot_file} ${lagrange_file} ${linear_combination_file}
+rm ${transcript_file}
+rm ${challenge_0_file}
+rm ${response_1_file} ${challenge_1_file} ${response_1_digest_file}
+rm ${response_2_file} ${challenge_2_file} ${response_2_digest_file}
+rm ${response_3_file} ${challenge_3_file} ${response_3_digest_file}
+invalid_response_digest_file=_test_response_digest_invalid.bin
+rm -f ${final_phase2_file}      # Same as challenge_3_file
```
AntoineRondelet: Let's remove the tabs/multiple spaces before the comment

src/mpc/cli/mpc_phase2_begin.cpp
```diff
@@ -0,0 +1,116 @@
...
+    std::string out_file;
+
+public:
+    mpc_phase2_begin() : subcommand("phase2-begin"), lin_comb_file(), out_file()
+    {
+    }
+
+private:
+    void initialize_suboptions(
+        boost::program_options::options_description &options,
```
AntoineRondelet: Why not use `po::` here?

src/mpc/cli/mpc_phase2_begin.cpp
```diff
@@ -0,0 +1,116 @@
...
+        const size_t num_inputs = [this]() {
+            libsnark::protoboard<FieldT> pb;
+            init_protoboard(pb);
+            return pb.num_inputs();
+        }();
+        libff::print_indent();
+        std::cout << std::to_string(num_inputs) << std::endl;
+        libff::leave_block("Computing num inputs");
+
+        // TODO: use the hash of the linear combination as the inital transcript
```
AntoineRondelet: What's the state on that  @dtebbs?

src/mpc/cli/mpc_phase2_verify_contribution.cpp
```diff
@@ -0,0 +1,151 @@
...
+        all_options.add(options).add_options()(
+            "challenge_file", po::value<std::string>(), "challenge file")(
+            "response_file", po::value<std::string>(), "response file");
+        pos.add("challenge_file", 1).add("response_file", 1);
+    }
+
+    void parse_suboptions(
+        const boost::program_options::variables_map &vm) override
+    {
+        if (0 == vm.count("challenge_file")) {
```
AntoineRondelet: `vm.count("challenge_file") == 0` (ie: switch left hand side and right hand side) or `if(!vm.count("challenge_file")){}`

src/snarks/groth16/mpc/chacha_rng.hpp
```diff
@@ -0,0 +1,29 @@
...
+#define __ZETH__SNARKS_GROTH16_MPC_CHACHA_RNG_HPP__
+
+#include <cstddef>
+#include <cstdint>
+
+namespace libzeth
+{
+
+// Random number generator matching the implementation used by zcash
+// powersoftau and phase2.
```
AntoineRondelet: Can you please provide a link to the document or source file that you follow?

src/util.cpp
```diff
@@ -213,7 +213,7 @@ std::string hexadecimal_str_to_binary_str(const std::string &s)
     return out;
 }
 
-std::string binary_str_to_hexadecimal_str(const char *s, const size_t size)
+std::string binary_str_to_hexadecimal_str(const void *s, const size_t size)
```
AntoineRondelet: Why change it to `void`?

src/test/powersoftau_test.cpp
```diff
@@ -51,12 +51,139 @@ template<typename T> std::string to_hex(const T &v)
     return bin_to_hex(bufstream.str());
 }
 
+template<typename T> using PointSerializer = void(std::ostream &, const T &);
```
AntoineRondelet: Did you use `CamelCase` for a specific reason? Otherwise, we may want to switch to `snake_case` to be consistent

scripts/ci
dtebbs: Changed the order in conditions

scripts/ci
dtebbs: Changed the order

scripts/ci
dtebbs: Changed the order

scripts/test_mpc
dtebbs: This, along with the next line, creates a digest file that is 4 zero bytes followed by the correct contribution hash (which, with high probability, will be read as a digest which is not in the transcript).  This checks that the `phase2-verify-transcript` command fails when used to check for a contribution that is not in the transcript.

scripts/test_mpc
dtebbs: Removed spaces before comment

src/mpc/cli/mpc_phase2_begin.cpp
dtebbs: switched to `po::`

src/snarks/groth16/mpc/chacha_rng.hpp
dtebbs: Added links to the usage and implementation

src/mpc/cli/mpc_phase2_begin.cpp
dtebbs: I recall mentioning this, but I do not recall any decision, so it's just left as a comment.  I'll make the change.  We will need to expose something for clients, so they can confirm the phase2 transcript begins with the expected digest.

src/mpc/cli/mpc_phase2_begin.cpp
AntoineRondelet: Thanks. Several other files had similar patterns though, have you changed all occurrences of that?

src/mpc/cli/mpc_phase2_contribute.cpp
```diff
@@ -0,0 +1,159 @@
...
+        , challenge_file()
+        , out_file()
+        , digest()
+        , skip_user_input(false)
+    {
+    }
+
+private:
+    void initialize_suboptions(
+        boost::program_options::options_description &options,
```
AntoineRondelet: Use `po::`

src/mpc/cli/mpc_phase2_verify_transcript.cpp
```diff
@@ -0,0 +1,201 @@
...
+        , challenge_0_file()
+        , transcript_file()
+        , final_challenge_file()
+        , digest()
+    {
+    }
+
+private:
+    void initialize_suboptions(
+        boost::program_options::options_description &options,
```
AntoineRondelet: `po::`

src/mpc/cli/mpc_phase2_begin.cpp
AntoineRondelet: Just checked and similar changes may need to be applied to other files. Just added other comments below to flag the locations where `po::` can be used.

src/mpc/cli/mpc_phase2_contribute.cpp
dtebbs: Replaced here and in all sucommands

src/mpc/cli/mpc_phase2_verify_transcript.cpp
dtebbs: Replaced here and in all subcommands

src/mpc/cli/mpc_phase2_begin.cpp
dtebbs: After discussing, I've added the following to this PR:
- added a `cs_hash` field to the phase2 accumulators
- this is set to the digest of the linear-combination when the initial challenge is created
- `verify_contribution` now checks that the cs_hash in response and challenge match
- `cs_hash` is used to "seed" the value of the transcript digest

src/mpc/cli/mpc_phase2_contribute.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+namespace
+{
+
+// Usage:
+//   $0 phase2-contribute [<options>] <challenge_file>
+//
+// Options:
+//   --out <file>        Response output file (mpc-response.bin)
+//   --digest <file>     Write contribution hash to file.
```
AntoineRondelet: Remove `.` at the end

src/mpc/cli/mpc_phase2_contribute.cpp
```diff
@@ -0,0 +1,158 @@
...
+namespace po = boost::program_options;
+
+namespace
+{
+
+// Usage:
+//   $0 phase2-contribute [<options>] <challenge_file>
+//
+// Options:
+//   --out <file>        Response output file (mpc-response.bin)
```
AntoineRondelet: Is `(mpc-response.bin)` the default filename?

src/mpc/cli/mpc_phase2_contribute.cpp
```diff
@@ -0,0 +1,158 @@
...
+//   --digest <file>     Write contribution hash to file.
+//   --skip-user-input   Use only ststem randomness
```
AntoineRondelet: ststem -> system

src/mpc/cli/mpc_phase2_contribute.cpp
```diff
@@ -0,0 +1,158 @@
...
+            std::cout << "digest: " << digest << std::endl;
+            std::cout << "skip_user_input: " << skip_user_input << std::endl;
+        }
+
+        libff::enter_block("Load challenge file");
+        srs_mpc_phase2_challenge<ppT> challenge = [&]() {
+            std::ifstream in(
+                challenge_file, std::ios_base::binary | std::ios_base::in);
+            return srs_mpc_phase2_challenge<ppT>::read(in);
+        }();
```
AntoineRondelet: Why not use `read_from_file` that you defined in `mpc_common.hpp`?

src/mpc/cli/mpc_phase2_contribute.cpp
```diff
@@ -0,0 +1,158 @@
...
+        libff::enter_block("Writing response");
+        libff::print_indent();
+        std::cout << out_file << std::endl;
+        {
+            std::ofstream out(out_file);
+            response.write(out);
+        }
+        libff::leave_block("Writing response");
+
+        srs_mpc_hash_t contr_digest;
```
AntoineRondelet: `contr_digest` -> `contrib_digest`

src/mpc/cli/mpc_phase2_verify_contribution.cpp
```diff
@@ -0,0 +1,150 @@
...
+        libff::leave_block("Load challenge file");
+
+        libff::enter_block("Load response file");
+        srs_mpc_phase2_response<ppT> response = [&]() {
+            std::ifstream in(
+                response_file, std::ios_base::binary | std::ios_base::in);
+            in.exceptions(
+                std::ios_base::eofbit | std::ios_base::badbit |
+                std::ios_base::failbit);
+            return srs_mpc_phase2_response<ppT>::read(in);
```
AntoineRondelet: Use `read_from_file` function?

src/mpc/cli/mpc_phase2_verify_transcript.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+        // Load the initial challenge
+        libff::enter_block("Load challenge_0 file");
+        const srs_mpc_phase2_challenge<ppT> challenge_0 = [&]() {
+            std::ifstream in(
+                challenge_0_file, std::ios_base::binary | std::ios_base::in);
+            in.exceptions(
+                std::ios_base::eofbit | std::ios_base::badbit |
+                std::ios_base::failbit);
+            return srs_mpc_phase2_challenge<ppT>::read(in);
```
AntoineRondelet: `read_from_file`?

src/mpc/cli/mpc_phase2_verify_transcript.cpp
```diff
@@ -0,0 +1,217 @@
...
+        }();
+        libff::leave_block("Load challenge_0 file");
+
+        // Simple sanity check on challenge.0.  The initial transcript digest
```
AntoineRondelet: `challenge.0.  The` -> double spacing. Please use a single space after `.`

src/snarks/groth16/mpc/chacha_rng.hpp
```diff
@@ -0,0 +1,35 @@
...
+// Usage:
+//   https://github.com/clearmatics/powersoftau
+//   (See hash_to_g2 function)
+// Implementation:
+//   https://github.com/rust-random/rand/blob/master/rand_chacha/src/chacha.rs
+//   (See description comment, in particular word layout)
```
AntoineRondelet: Let's use `///` type of comments as this is a useful comment for the documentation

src/snarks/groth16/mpc/hash_utils.hpp
```diff
@@ -0,0 +1,91 @@
...
+#define __ZETH_SNARKS_GROTH16_MPC_HASH_UTILS_HPP__
+
+#include <ios>
+#include <iostream>
+#include <sodium/crypto_generichash_blake2b.h>
+
+namespace libzeth
+{
+
+// Hashing for MPC.  Streaming and whole-buffer interfaces.
```
AntoineRondelet: Spacing after `.`

src/snarks/groth16/mpc/phase2.cpp
```diff
@@ -0,0 +1,66 @@
...
+    libff::alt_bn128_G2_write_compressed(out, delta_g2);
+    for (const G1 h : H_g1) {
+        libff::alt_bn128_G1_write_compressed(out, h);
+    }
+    for (const G1 l : L_g1) {
+        libff::alt_bn128_G1_write_compressed(out, l);
+    }
+}
+
+template<>
```
AntoineRondelet: What's the role of this?

src/snarks/groth16/mpc/phase2.hpp
```diff
@@ -0,0 +1,309 @@
...
+template<typename ppT> class srs_powersoftau;
+template<typename ppT> class srs_mpc_layer_L1;
+
+/// Target of the MPC for Phase2 of the SRS generation.  Follows exactly $M_2$
+/// in section 7.3 of [BoweGM17], whre we use L in place of K, consistent with
+/// the keypair in libsnark.
+template<typename ppT> class srs_mpc_phase2_accumulator
+{
+public:
+    // Hash of the initial state before any contributions are made.  Kept
```
AntoineRondelet: Spacing after `.`

src/snarks/groth16/mpc/phase2.hpp
```diff
@@ -0,0 +1,309 @@
...
+    // constant over the MPC, and used to check that challenges and responses
+    // are part of the same MPC.
+    srs_mpc_hash_t cs_hash;
+
+    libff::G1<ppT> delta_g1;
+
+    libff::G2<ppT> delta_g2;
+
+    // { H_i } = { [ t(x) . x^i / delta ]_1 }  i=0..n-2 (n-1 entries)
```
AntoineRondelet: `i=0..n-2 (n-1 entries)` => `i \in [0..n-2] (n-1 entries)` likewise for `j` below

src/snarks/groth16/mpc/phase2.hpp
```diff
@@ -0,0 +1,309 @@
...
+
+    bool operator==(const srs_mpc_phase2_accumulator<ppT> &other) const;
+    bool is_well_formed() const;
+    void write(std::ostream &out) const;
+    void write_compressed(std::ostream &out) const;
+    static srs_mpc_phase2_accumulator<ppT> read(std::istream &in);
+    static srs_mpc_phase2_accumulator<ppT> read_compressed(std::istream &in);
+};
+
+/// Public Key representing a single contribution to the MPC.  This includes a
```
AntoineRondelet: Spacing after `.`

src/snarks/groth16/mpc/phase2.hpp
```diff
@@ -0,0 +1,309 @@
...
+/// proof-of-knowledge, consisting of some point $s$ in G1, chosen randomly, $s
+/// * \delta$, $r$ generated deterministically by hashing $s$ and $s * \delta$
+/// into G2, and $r * \delta$.
```
AntoineRondelet: The transcript should also show up inside the `hash_to_G2` when computing `r`, right?

src/snarks/groth16/mpc/phase2.hpp
```diff
@@ -0,0 +1,309 @@
...
+void srs_mpc_digest_to_fp(
+    const srs_mpc_hash_t transcript_digest,
+    libff::Fp_model<n, modulus> &out_fr);
+
+template<typename ppT>
+libff::G2<ppT> srs_mpc_digest_to_g2(const srs_mpc_hash_t digest);
+
+/// Given the output from the linear combination of the L1 layer of the SRS
+/// circuit, compute the starting parameters for Phase 2 (the MPC for C2
+/// layer).  See "Initialization" in section 7.3 of [BoweGM17].
```
AntoineRondelet: Spacing after `.` (applicable to other places below)

src/snarks/groth16/mpc/mpc_utils.tcc
```diff
@@ -0,0 +1,257 @@
...
+           libzeth::container_is_well_formed(B_g2) &&
+           libzeth::container_is_well_formed(ABC_g1);
+}
+
+template<typename ppT>
+void srs_mpc_layer_L1<ppT>::write(std::ostream &out) const
+{
+    using G1 = libff::G1<ppT>;
+    using G2 = libff::G2<ppT>;
+    check_well_formed(*this, "mpc_layer1 (write)");
```
AntoineRondelet: In `util.tcc` there are 2 functions:
```cpp
template<typename T> void check_well_formed(const T &v, const char *name)
{
    if (!v.is_well_formed()) {
        throw std::invalid_argument(std::string(name) + " not well-formed");
    }
}

template<typename T> void check_well_formed_(const T &v, const char *name)
{
    if (!is_well_formed(v)) {
        throw std::invalid_argument(std::string(name) + " not well-formed");
    }
}
```

Why having two very similar functions like that?
What about having a template `StructuredT` to represent structured types for which it makes sense to define a `check_well_formed` method? We can document the functions that are expected to be implemented by a `StructuredT` type to have a sort of interface. Otherwise, it may be worth using abstract classes for that, and use inheritance for all the objects that need to implement a `check_well_formed` method. Doing so makes the code base clearer (seeing a class inheriting from `structured` being an abstract class makes the code base easier to read and clearer as the interfaces are well defined and easier to maintain IMO)

src/snarks/groth16/mpc/phase2.cpp
```diff
@@ -0,0 +1,66 @@
...
+{
+
+template<>
+void srs_mpc_phase2_accumulator<libff::alt_bn128_pp>::write_compressed(
+    std::ostream &out) const
+{
+    using G1 = libff::alt_bn128_G1;
+    check_well_formed(*this, "mpc_layer2 (write)");
+
+    // Write fixed-sizes cs_hash and sizes first.
```
AntoineRondelet: `Write fixed-sizes cs_hash` -> Not sure I understand. What do you mean by "fixed-sizes cs_hash"?

src/snarks/groth16/mpc/phase2.hpp
```diff
@@ -0,0 +1,309 @@
...
+/// Outputs the public key (which includes the POK) for our secret.  Correponds
+/// to steps 1 and 2 in "Computation", section 7.3 of [BoweGM17]
+template<typename ppT>
+srs_mpc_phase2_publickey<ppT> srs_mpc_phase2_compute_public_key(
+    const srs_mpc_hash_t transcript_digest,
+    const libff::G1<ppT> &last_delta,
+    const libff::Fr<ppT> &secret);
+
+/// Verifies that a public key is correct, given some previous delta.
+/// Corresponds to steps 1 and 2 from "Verfication" in section 7.3 of
```
AntoineRondelet: `Verfication` -> Verification

src/snarks/groth16/mpc/phase2.hpp
```diff
@@ -0,0 +1,309 @@
...
+/// Core update function, which applies a secret contribution to an
+/// accumulator.  Corresponds to steps 3 onwards in "Computation",
+/// section 7.3 of [BoweGM17].
+template<typename ppT>
+srs_mpc_phase2_accumulator<ppT> srs_mpc_phase2_update_accumulator(
+    const srs_mpc_phase2_accumulator<ppT> &last_accum,
+    const libff::Fr<ppT> &delta_j);
+
+/// Assuming last is fully verified, and updated.delta_g1 has the appropriate
+/// ratio, check that all other elements of updated.  This covers the G2 part
```
AntoineRondelet: `check that all other elements of updated.` -> Unfinished sentence?

src/snarks/groth16/mpc/phase2.hpp
```diff
@@ -0,0 +1,309 @@
...
+/// consistency of initial and final accumulators without intermediate values.
+template<typename ppT>
+bool srs_mpc_phase2_update_is_consistent(
+    const srs_mpc_phase2_accumulator<ppT> &last,
+    const srs_mpc_phase2_accumulator<ppT> &updated);
+
+/// Core verification function for a single contribution.  Checks the
+/// self-consistency of a public key, and that the corresponding contribution
+/// has been correctly applied to all values in 'last', to generate 'updated'.
+/// This corresponds to "Verfication" in section 7.3 of [BoweGM17] (for a
```
AntoineRondelet: `Verfication` -> Verification

src/snarks/groth16/mpc/phase2.hpp
```diff
@@ -0,0 +1,309 @@
...
+/// substitute for the full MPC with an auditable log of
+/// contributions, but is useful for testing.
+template<typename ppT>
+srs_mpc_phase2_challenge<ppT> srs_mpc_dummy_phase2(
+    const srs_mpc_layer_L1<ppT> &layer1,
+    const libff::Fr<ppT> &delta,
+    size_t num_inputs);
+
+/// Given the output from all phases of the MPC, create the
+/// prover and verification keys for the given circuit.
```
AntoineRondelet: `prover and verification keys` -> "proving and verification keys", or "prover and verifier keys"

src/snarks/groth16/mpc/phase2.tcc
```diff
@@ -0,0 +1,794 @@
...
+    return !memcmp(cs_hash, other.cs_hash, sizeof(srs_mpc_hash_t)) &&
+           (delta_g1 == other.delta_g1) && (delta_g2 == other.delta_g2) &&
+           (H_g1 == other.H_g1) && (L_g1 == other.L_g1);
+}
+
+template<typename ppT>
+bool srs_mpc_phase2_accumulator<ppT>::is_well_formed() const
+{
+    return delta_g1.is_well_formed() && delta_g2.is_well_formed() &&
+           libzeth::container_is_well_formed(H_g1) &&
```
AntoineRondelet: `libzeth::`, you're in the `libzeth` namespace, do you really need this prefix?

src/snarks/groth16/mpc/phase2.tcc
```diff
@@ -0,0 +1,794 @@
...
+    const libff::Fr<ppT> &delta_j)
+{
+    libff::enter_block("call to srs_mpc_phase2_update_accumulator");
+    const libff::Fr<ppT> delta_j_inverse = delta_j.inverse();
+
+    // Step 3 (from [BoweGM17]): Update accumulated $\delta$
+    const libff::G1<ppT> new_delta_g1 = delta_j * last_accum.delta_g1;
+    const libff::G2<ppT> new_delta_g2 = delta_j * last_accum.delta_g2;
+
+    // Step3: Update $L_i$ by dividing by $\delta$ ('K' in the paper, but we
```
AntoineRondelet: `Step3` -> `Step 3`

src/snarks/groth16/mpc/phase2.tcc
```diff
@@ -0,0 +1,794 @@
...
+    return srs_mpc_phase2_response<ppT>(
+        std::move(new_accum), std::move(pubkey));
+}
+
+template<typename ppT>
+bool srs_mpc_phase2_verify_response(
+    const srs_mpc_phase2_challenge<ppT> &challenge,
+    const srs_mpc_phase2_response<ppT> &response)
+{
+    // Ensure that response.pubkey corresponsds to challenge.transcript_digest
```
AntoineRondelet: `corresponsds` -> corresponds

src/mpc/cli/mpc_phase2_contribute.cpp
dtebbs: Yes, exactly.  I think this is a fairly common way to express it.  Let me know if you prefer some other way.

src/snarks/groth16/mpc/phase2.cpp
dtebbs: This provides the implementation for this specific value of `ppT` (we can only choose to write compressed / uncompressed points for `ppT == alt_bn128_pp`, so we can't provide an implementation that is generic in `ppT` without changing libff to support this for all curves)

dtebbs: (To clarify, by default compression in libff is statically set with a `#define` controlled by the build config.  The functions such as `libff::alt_bn128_G1_write_compressed` were added by me in order to allow us to use both the compressed and uncompressed formats in a single build).

src/snarks/groth16/mpc/phase2.cpp
dtebbs: Sorry, that was a typo ... should have been `fixed-size cs_hash`.  I've reworded the comment.

src/snarks/groth16/mpc/phase2.hpp
dtebbs: Removed space

src/snarks/groth16/mpc/phase2.hpp
dtebbs: I've added a mention of the transcript digest.

src/snarks/groth16/mpc/phase2.hpp
dtebbs: changed

src/snarks/groth16/mpc/phase2.hpp
dtebbs: changed

src/snarks/groth16/mpc/phase2.hpp
dtebbs: changed

src/snarks/groth16/mpc/hash_utils.hpp
dtebbs: removed the space

src/mpc/cli/mpc_phase2_verify_transcript.cpp
dtebbs: Removed

src/snarks/groth16/mpc/chacha_rng.hpp
dtebbs: Switched to `///`

src/mpc/cli/mpc_phase2_verify_transcript.cpp
dtebbs: Used `read_from_file`

src/mpc/cli/mpc_phase2_verify_contribution.cpp
dtebbs: Changed

src/mpc/cli/mpc_phase2_contribute.cpp
dtebbs: CHanged to use `read_from_file`

src/mpc/cli/mpc_phase2_contribute.cpp
dtebbs: Fixed

src/mpc/cli/mpc_phase2_contribute.cpp
dtebbs: Removed .

src/mpc/cli/mpc_phase2_begin.cpp
```diff
@@ -0,0 +1,119 @@
...
+        libff::enter_block("Load linear combination file");
+        srs_mpc_hash_t cs_hash;
+        srs_mpc_layer_L1<ppT> lin_comb = [&]() {
+            std::ifstream inf(
+                lin_comb_file, std::ios_base::binary | std::ios_base::in);
+            hash_istream_wrapper in(inf);
+            srs_mpc_layer_L1<ppT> lin_comb = srs_mpc_layer_L1<ppT>::read(in);
+            in.get_hash(cs_hash);
+            return lin_comb;
+        }();
```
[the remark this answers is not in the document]
dtebbs: The `hash_istream_wrapper` wraps the istream and computes the hash while loading (to save having to reserialize the struct or reload the file.  I've removed the lambda by adding another util fucntion `read_from_file_and_hash` for this situation.

src/mpc/cli/mpc_phase2_contribute.cpp
dtebbs: Renamed

src/snarks/groth16/mpc/phase2.hpp
dtebbs: fixed typo

src/snarks/groth16/mpc/phase2.hpp
dtebbs: Fixed sentence

src/snarks/groth16/mpc/phase2.hpp
dtebbs: updated

src/snarks/groth16/mpc/phase2.tcc
dtebbs: removed namespace prefix

src/snarks/groth16/mpc/phase2.tcc
dtebbs: Fixed

src/snarks/groth16/mpc/phase2.tcc
dtebbs: Fixed

src/snarks/groth16/mpc/mpc_utils.tcc
dtebbs: For all types we define ourselves (where it makes sense) we include the `is_well_formed` method, so `check_well_formed` can be used.
`check_well_formed_` is for the case where we want to perform this kind of check on a type which is not part of zeth (I think mainly the proving and verification keys in libsnark), where we cannot easily add a new method.

(I am not sure if this has answered your comment here - let me know if I've misunderstood the point.)

src/mpc/cli/mpc_phase2_contribute.cpp
AntoineRondelet: I'd write something like: `--out <file>        Response output file (default "mpc-response.bin")`

In addition to that; I'd be in favor to use environment variables to avoid creating files in the location where the command is ran. Maybe we can use the env var `ZETH_TRUSTED_SETUP_DIR` for that, or just $HOME

src/snarks/groth16/mpc/mpc_utils.tcc
AntoineRondelet: I see, thanks for the clarification. Maybe we can add a comment to reflect your explanation above into the code?

> (I am not sure if this has answered your comment here - let me know if I've misunderstood the point.)

Yes you answered my first question, thanks.

I still think that using inherited abstract classes can be useful to define our interfaces. What do you think?

dtebbs: (To capture the results of a discussion ...) I agree that the abstract base classes approach will better capture these constraints.  We agreed to think about it.

For now, the point about naming seems a good idea and minimal impact.  I've renamed this (and type params in other places) to be more meaningful than just `T`.

dtebbs:
> So, is this TODO still worth keeping?
Removed.  Thanks.

src/mpc/cli/mpc_phase2_begin.cpp
dtebbs: The TODO should have been removed in the latest version.

src/mpc/cli/mpc_phase2_begin.cpp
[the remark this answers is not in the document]
AntoineRondelet: I see, thanks

src/snarks/groth16/mpc/mpc_utils.tcc
AntoineRondelet: Thanks, please add comments where relevant to clarify what "interfaces" each type implement (eg: https://github.com/clearmatics/zeth/blob/develop/src/circuits/sha256/sha256_ethereum.hpp#L8)

src/mpc/cli/mpc_phase2_contribute.cpp
dtebbs: Based on our discussion, I've removed these --out flags from all the mpc subcommands, so the user is forced to specify the output file explicitly.

src/snarks/groth16/mpc/mpc_utils.tcc
dtebbs: Thanks.  I've added comments describing the new `StructuredT` and `ReadableT` "interfaces" and a comment on each strcture that implements them.

src/mpc/cli/mpc_common.cpp
```diff
@@ -0,0 +1,177 @@
...
+}
+
+int mpc_main(
+    int argc,
+    char **argv,
+    const std::map<std::string, subcommand *> &commands,
+    ProtoboardInitFn pb_init)
+{
+    ppT::init_public_params();
+    po::options_description global("");
```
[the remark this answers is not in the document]
dtebbs: Good idea.  Added.

src/snarks/groth16/mpc/powersoftau_utils.hpp
```diff
@@ -8,9 +8,9 @@
 namespace libzeth
 {
 
-/// Output from the first phase of the MPC (powersoftau).  The
-/// structure matches that data exactly (no indication of degree,
-/// etc), so that it can be loaded directly.
+/// Output from the first phase of the MPC (powersoftau).  The structure
```
AntoineRondelet: Spacing

src/mpc/cli/mpc_common.hpp
```diff
@@ -0,0 +1,95 @@
+#ifndef __ZETH_MPC_MPC_COMMON_HPP__
+#define __ZETH_MPC_MPC_COMMON_HPP__
```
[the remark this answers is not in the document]
dtebbs: Updated.

src/mpc/cli/mpc_phase2_verify_contribution.cpp
```diff
@@ -0,0 +1,140 @@
...
+            srs_mpc_phase2_verify_response(challenge, response);
+        libff::leave_block("Verifying response");
+        if (!response_is_valid) {
+            std::cerr << "Response is invalid" << std::endl;
+            return 1;
+        }
+
+        // TODO: Backup the transcript file before writing a new version?
+
+        // If a transcation file has been specified, append this contribution
```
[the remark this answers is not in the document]
dtebbs: changed to `transcript`

src/snarks/groth16/mpc/chacha_rng.hpp
```diff
@@ -0,0 +1,35 @@
+#ifndef __ZETH__SNARKS_GROTH16_MPC_CHACHA_RNG_HPP__
```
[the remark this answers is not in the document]
dtebbs: Fixed

src/snarks/groth16/mpc/mpc_utils.hpp
```diff
@@ -27,8 +27,9 @@ namespace libzeth
 template<typename ppT> class srs_powersoftau;
 template<typename ppT> class srs_lagrange_evaluations;
 
-/// Output from linear combination $L_1$ - the linear combination of
-/// elements in powersoftau, based on a specific circuit.
+/// Output from linear combination $L_1$ - the linear combination of elements
+/// in powersoftau, based on a specific circuit.  Implements the interfaces of
```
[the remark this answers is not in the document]
dtebbs: changed and replaced all others in file.

src/mpc/cli/mpc_phase2_verify_contribution.cpp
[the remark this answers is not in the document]
AntoineRondelet: Ah yes indeed, wasn't paying attention here

src/snarks/groth16/mpc/hash_utils.hpp
```diff
@@ -0,0 +1,91 @@
...
+void srs_mpc_hash_final(srs_mpc_hash_state_t &, srs_mpc_hash_t);
+void srs_mpc_compute_hash(
+    srs_mpc_hash_t out_hash, const void *data, size_t data_size);
+void srs_mpc_compute_hash(srs_mpc_hash_t out_hash, const std::string &data);
+
+/// Convert a hash to a human-readable string (4 x 4 x 4-byte hex words),
+/// following the format used in the "powersoftau" and "Sapling MPC" code.
+void srs_mpc_hash_write(const srs_mpc_hash_t hash, std::ostream &out);
+
+/// Parse a human-readable string (4 x 4 x 4-byte hex words) reprsenting an
```
AntoineRondelet: reprsenting -> representing

src/snarks/groth16/mpc/hash_utils.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+#include <ios>
+#include <iostream>
+#include <sodium/crypto_generichash_blake2b.h>
+
+namespace libzeth
+{
+
+// Hashing for MPC. Streaming and whole-buffer interfaces.
+const size_t srs_mpc_hash_size = 64;
```
AntoineRondelet: Use uppercase for consts `srs_mpc_hash_size` -> `SRS_MPC_HASH_SIZE`

AntoineRondelet: (applicable to other consts on this PR)

src/snarks/groth16/mpc/hash_utils.hpp
```diff
@@ -0,0 +1,91 @@
...
+const size_t srs_mpc_hash_array_length = srs_mpc_hash_size / sizeof(size_t);
+using srs_mpc_hash_t = size_t[srs_mpc_hash_array_length];
```
AntoineRondelet: Group `using` together, and `const` together

src/mpc/cli/mpc_common.cpp
```diff
@@ -0,0 +1,177 @@
...
+        std::cout << "Usage:\n"
+                  << "  " << argv[0]
+                  << " [<options>] <command> <command-arguments> ...\n\n"
+                  << global;
+
+        std::cout << "\nCommands:\n";
+        list_commands(commands);
+        // for (const auto &cmd : commands) {
+        //     std::cout << "  " << cmd.first;
+        // }
```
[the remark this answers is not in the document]
dtebbs: Removed.  Thanks.

src/test/powersoftau_test.cpp
```diff
@@ -51,12 +51,139 @@ template<typename T> std::string to_hex(const T &v)
     return bin_to_hex(bufstream.str());
 }
 
+template<typename T> using point_serializer_t = void(std::ostream &, const T &);
+
+template<typename T> using point_deserializer_t = void(std::istream &, T &);
+
+template<
+    typename T,
+    size_t expectSize,
```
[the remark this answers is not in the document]
dtebbs: renamed the variable

src/snarks/groth16/mpc/hash_utils.hpp
dtebbs: fixed spelling.

src/snarks/groth16/mpc/hash_utils.hpp
dtebbs: renamed all constants to upper case

src/snarks/groth16/mpc/hash_utils.hpp
dtebbs: grouped

src/snarks/groth16/mpc/chacha_rng.hpp
```diff
@@ -0,0 +1,35 @@
...
+{
+public:
+    chacha_rng(const void *seed, size_t seed_size);
+    void random(void *output, size_t output_size);
+
+private:
+    void populate();
+
+    uint32_t key[8];
+    uint32_t iv[4];
```
AntoineRondelet: The IV should be 3 words, not 4

src/snarks/groth16/mpc/chacha_rng.hpp
```diff
@@ -0,0 +1,35 @@
...
+    uint8_t data[64];
```
AntoineRondelet: You mean, block here, right?

src/snarks/groth16/mpc/chacha_rng.cpp
```diff
@@ -0,0 +1,65 @@
...
+{
+
+chacha_rng::chacha_rng(const void *seed, size_t seed_size)
+    : data_used(sizeof(data))
+{
+    seed_size = std::min(seed_size, sizeof(key));
+    memcpy(key, seed, seed_size);
+    if (seed_size < sizeof(key)) {
+        memset(((uint8_t *)key) + seed_size, 0, sizeof(key) - seed_size);
+    }
```
AntoineRondelet: Here, you follow the `from_seed` function  https://github.com/rust-random/rand/blob/master/rand_chacha/src/chacha.rs#L144, right?

dtebbs:
> > `template<>`

> Let's either add a comment if necessary rather than using an empty template. Templatized code aims to be written in a tcc file anyway.

(@AntoineRondelet for some reason, github is not letting me reply directly, so I've had to move this to a new comment).

I've added some comments to explain these functions and why they are here.  Just in case it's not clear, this code will only work for the case where `ppT == libff::alt_bn128_pp`, and the empty template here is to specify that this is a concrete function for the case `ppT == libff::alt_bn128_pp`, rather than generic templated code.

Since this is a concrete function, if it was in a .tcc file it would likely cause link errors due to the symbol being defined multiple times.  If we turned it into fully generic code with `template<typename ppT>` then it would not compile for any other `ppT`, and it would make it difficult to provide implementations for other curves in the future.

Let me know if the comments are not clear, or you think we should take a different approach.

AntoineRondelet:
> > > `template<>`
> 
> > Let's either add a comment if necessary rather than using an empty template. Templatized code aims to be written in a tcc file anyway.
> 
> (@AntoineRondelet for some reason, github is not letting me reply directly, so I've had to move this to a new comment).
> 
> I've added some comments to explain these functions and why they are here. Just in case it's not clear, this code will only work for the case where `ppT == libff::alt_bn128_pp`, and the empty template here is to specify that this is a concrete function for the case `ppT == libff::alt_bn128_pp`, rather than generic templated code.
> 
> Since this is a concrete function, if it was in a .tcc file it would likely cause link errors due to the symbol being defined multiple times. If we turned it into fully generic code with `template<typename ppT>` then it would not compile for any other `ppT`, and it would make it difficult to provide implementations for other curves in the future.
> 
> Let me know if the comments are not clear, or you think we should take a different approach.

Yes. What I was saying was basically: if this is not generic code, then let's remove `template<>`, and add a comment saying "this code will only work for the case where `ppT == libff::alt_bn128_pp`"

src/snarks/groth16/mpc/chacha_rng.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+void chacha_rng::populate()
+{
+    memset(data, 0, sizeof(data));
+    crypto_stream_chacha20_ietf_xor_ic(
+        (uint8_t *)data,
+        (uint8_t *)data,
+        sizeof(data),
+        (const uint8_t *)&iv[1],
+        iv[0],
```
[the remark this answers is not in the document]
AntoineRondelet: Do you really need to use `crypto_stream_chacha20_ietf_xor_ic` and not `crypto_stream_chacha20_ietf_xor`? I don't think you are setting the initial value of the block counter to a non-zero value, do you?

src/snarks/groth16/mpc/phase2.cpp
AntoineRondelet: I though we could just remove `template<>` because the type was specified, but the compiler is not happy with that, so let's keep it then.

src/snarks/groth16/mpc/chacha_rng.hpp
dtebbs: Sorry, it was badly named.  I've renamed this to `counter` to match the rust implementation.

src/snarks/groth16/mpc/chacha_rng.hpp
dtebbs: Renamed to `block`

src/snarks/groth16/mpc/chacha_rng.cpp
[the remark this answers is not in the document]
dtebbs: I've added comments to try and clarify how the words in `iv` (now called `counter`) correspond to these parameters.  Let me know if they are not clear and I'll try to expand.

src/snarks/groth16/mpc/chacha_rng.cpp
dtebbs: I've added comments here and in the other functions referring to the rust implementation.  Also renamed the internal `populate` method to match `update` in rust.

AntoineRondelet: Perfect, thanks

src/snarks/groth16/mpc/chacha_rng.hpp
AntoineRondelet: thanks

AntoineRondelet:
LGTM

Issue #52, opened 2019-08-16
Modify the packing policy to reduce the number of primary inputs processed by the Verifier


AntoineRondelet:
In the current state of the project, several primary inputs are digest (of bit-length 256) and are packed into field elements. This set of field elements represent the set of primary inputs sent to the Verifier (contract), which then executes the SNARK verification routine. The verification routine of the SNARK used does a number of scalar multiplications linear in the number of primary inputs.

For now, each 256-bit digest (`d`) is packed into 2 field elements:
1. `d1`: One field element containing 253 bits from the digest
2. `d2`: Another field element containing the 3 remaining bits of the digest (and all zeroes afterwards)
This "packing policy" was chosen to keep things simple, but we clearly see that, if we have the following set of primary inputs:
`{rt, n1, n2, c1, c2, vpub_in, vpub_out}`, where `{n1,n2,c1,c2}` are digests, then, we'll send the set `{rt, n11, n12, n21, n22, c11, c12, c21, c22, vpub_in, vpub_out}` of field elements as the set of primary inputs, to the Verifier.

This is quite inefficient since we know that `{n12, n22, c12, c22}` will only contain 3 "meaningful bits" (the other bits in the field element representation will be set to 0's).

It'd be better to represent the information of the set `{n12, n22, c12, c22}` in a single field element.

Doing so would reduce the cardinality of the set of primary inputs and save a few calls to the `bn256ScalarMul` precompiled contract, saving gas on the Verifier side.

AntoineRondelet:
This requires changes in the circuit (to change the way we pack the primary inputs) and changes on the solidity side (to parse and recompose the digests from the packed elements - some "utils" functions will need to be modified/implemented)

rrtoledo:
Issue #55 can be incorporated here.

rrtoledo:
Could we go a step further in the packing ?
`v_in` and `v_out` each need only 64 bits. A field variable being written over 253 bits, we could pack `v_in`, `v_out` and the `{n12, n22, c12, c22, hsig2, h02, h12}` into  a single field variable (2*64+3*(2* NumInput + 1 + NumOutput) = 146 bits ).

AntoineRondelet:
Yes, I think this is fine to include the `v_in` and `v_out` values in the resulting field element @rrtoledo 
You'll need to make sure you can extract these values on the contract (like for other public inputs) to carry out all the relevant checks.

AntoineRondelet:
[...]

PR #80, opened 2019-09-24
Packing

rrtoledo:
This PR tackles issue #52 .


src/circuits/joinsplit.tcc
```diff
@@ -21,64 +20,64 @@ template<
...
+     *little bit the verification time) but this makes easier to retrieve the
+     *root and each nullifiers from the public inputs
+     **/
     std::array<
         std::shared_ptr<libsnark::multipacking_gadget<FieldT>>,
-        NumInputs + NumOutputs + 1 + 1 + 1 + NumInputs>
+        NumInputs + NumOutputs + 1 + NumInputs + 1>
         packers;
 
+    // TODO: Remove ZERO and pass it in the constructor
```
AntoineRondelet: This is not on develop, I guess this branch has forked some time ago from develop. Please make sure to apply your changes on top of the latest changes on develop to avoid re-introducing obsolete changes

src/circuits/joinsplit.tcc
```diff
@@ -115,75 +114,76 @@ public:
...
             for (size_t i = NumInputs; i < NumInputs + NumOutputs; i++) {
                 packed_inputs[i].allocate(
                     pb,
-                    1 + 1,
+                    1,
                     FMT(this->annotation_prefix, " out_commitment[%zu]", i));
             }
 
-            // We allocate 1 field element to pack the value (v_pub_in)
+            // We allocate a field elements to pack the value of h_sig
```
AntoineRondelet: `a field elements` -> a field element

src/circuits/joinsplit.tcc
```diff
@@ -21,64 +20,64 @@ template<
...
+     * Multipacking gadgets for the inputs (nullifierS, commitmentS, val_pub_in,
+     *val_pub_out) (the root is a field element) `NumInputs + NumOutputs`
+     *because we pack the nullifiers (Inputs of JS = NumInputs), the commitments
+     *(Output of JS = NumOutputs) AND the v_pub_out taken out of the mix (+1)
+     *AND the public value v_pub_in that is put into the mix (+1) AND the
+     *signature hash h_sig (+1) AND the malleability tags h_iS (+ NumInputs)
+     **/
+    size_t bits = 2 * 64 + 3 * (1 + 2 * NumInputs + NumOutputs);
+    size_t log_r =
+        253; // log_2(21888242871839275222246405745257275088548364400416034343698204186575808495617);
```
AntoineRondelet: Comment above the line please to avoid such line break when running clang-format

src/circuits/joinsplit.tcc
```diff
@@ -21,64 +20,64 @@ template<
...
-    // NumInputs)
+    /*
+     * Multipacking gadgets for the inputs (nullifierS, commitmentS, val_pub_in,
+     *val_pub_out) (the root is a field element) `NumInputs + NumOutputs`
+     *because we pack the nullifiers (Inputs of JS = NumInputs), the commitments
+     *(Output of JS = NumOutputs) AND the v_pub_out taken out of the mix (+1)
+     *AND the public value v_pub_in that is put into the mix (+1) AND the
+     *signature hash h_sig (+1) AND the malleability tags h_iS (+ NumInputs)
+     **/
+    size_t bits = 2 * 64 + 3 * (1 + 2 * NumInputs + NumOutputs);
```
AntoineRondelet: Please comment the purpose of this variable (ie: residual bits from the packing of 256 bit string into 253-bit long field elements), and rename to something more indicative like `residual_bits` or `residual_packing_bits`

src/circuits/joinsplit.tcc
```diff
@@ -21,64 +20,64 @@ template<
...
 
-    // We use an array of multipackers here instead of a single packer that
-    // packs everything. This leads to more public inputs (and thus affects a
-    // little bit the verification time) but this makes easier to retrieve the
-    // root and each nullifiers from the public inputs
+    /* We use an array of multipackers here instead of a single packer that
+     *packs everything. This leads to more public inputs (and thus affects a
+     *little bit the verification time) but this makes easier to retrieve the
+     *root and each nullifiers from the public inputs
+     **/
```
AntoineRondelet: The comment above should be modified accordingly

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -57,9 +56,15 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
     uint constant jsOut = 2; // Nb of commitments/ciphertexts
 
     // We have 2 field elements for each digest (nullifierS (jsIn), commitmentS (jsOut), h_iS (jsIn) and h_sig)
-    // The root, v_pub_in and v_pub_out are all represented by one field element, so we have 1 + 1 + 1 extra public values
+    // The root, v_pub_in and v_pub_out are all represented by one field element, so we have 1 extra public values
     uint constant nbInputs = 1 + 2 * (jsIn + jsOut) + 1 + 1 + 2 * (1 + jsIn);
 
+    // Constants regarding the hash digest length, the prime number used and its associated length in bits and the max values (v_in and v_out)
+    uint constant digest_length = 256;
+    uint r = 21888242871839275222246405745257275088548364400416034343698204186575808495617;
```
AntoineRondelet: `r` is not used. Only its bit length should be required to get the bit-length of scalar field elements and parse the public inputs received here.

src/circuits/joinsplit.tcc
```diff
@@ -21,35 +20,44 @@ template<
...
-        NumInputs + NumOutputs + 1 + 1 + 1 + NumInputs>
+        NumInputs + NumOutputs + 1 + NumInputs + 1>
         unpacked_inputs;
+    // Residual bits from packing of 256-bit long string into 253-bit long field
+    // elements to which are added the public value of size 64 bits
+    const size_t residual_packing_bits =
+        2 * 64 + 3 * (1 + 2 * NumInputs + NumOutputs);
+    // r =
+    // 21888242871839275222246405745257275088548364400416034343698204186575808495617
+    const size_t log_r = 253;
```
AntoineRondelet: Why not use `FieldT::capacity()`?

src/circuits/joinsplit.tcc
```diff
@@ -118,60 +126,59 @@ public:
...
                 packed_inputs[i].allocate(
-                    pb, 1 + 1, FMT(this->annotation_prefix, " h_i[%zu]", i));
+                    pb, 1, FMT(this->annotation_prefix, " h_i[%zu]", i));
             }
 
+            // We allocate as many field elements as needed to pack the public
+            // values and unpacked residual bits
+            packed_inputs[NumInputs + NumOutputs + 1 + NumInputs].allocate(
+                pb,
+                (residual_packing_bits + log_r - 1) / log_r,
```
AntoineRondelet: Why not use `libff::div_ceil` and the field capacity `FieldT::capacity`?
ie: `libff::div_ceil(residual_packing_bits, FieldT::capacity());`

src/circuits/joinsplit.tcc
```diff
@@ -213,13 +221,19 @@ public:
...
+                pb, 64, FMT(this->annotation_prefix, " zk_vpub_out"));
+
             // Initialize the unpacked input corresponding to the input
             // NullifierS
             for (size_t i = 0; i < NumInputs; i++) {
                 unpacked_inputs[i].insert(
                     unpacked_inputs[i].end(),
                     input_nullifiers[i]->bits.begin(),
-                    input_nullifiers[i]->bits.end());
+                    input_nullifiers[i]->bits.begin() + log_r);
```
AntoineRondelet: FieldT::capacity

src/circuits/joinsplit.tcc
```diff
@@ -598,32 +646,32 @@ public:
         nb_elements += 1;
 
         // Each nullifier is represented by 2 field elements (if we consider a
-        // digest_len of 256 bits)
+        // digest_len of 256 bits), including the "blob"
```
AntoineRondelet: What is "the blob"?

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -57,9 +56,16 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
     uint constant jsOut = 2; // Nb of commitments/ciphertexts
 
     // We have 2 field elements for each digest (nullifierS (jsIn), commitmentS (jsOut), h_iS (jsIn) and h_sig)
-    // The root, v_pub_in and v_pub_out are all represented by one field element, so we have 1 + 1 + 1 extra public values
+    // The root, v_pub_in and v_pub_out are all represented by one field element, so we have 1 extra public values
```
AntoineRondelet: `v_pub_in` and `v_pub_out` are not represented by a field element anymore. Update comments accordingly to the changes you've made

src/circuits/joinsplit.tcc
```diff
@@ -21,35 +20,44 @@ template<
...
+        NumInputs + NumOutputs + 1 + NumInputs + 1>
         packed_inputs;
     std::array<
         libsnark::pb_variable_array<FieldT>,
-        NumInputs + NumOutputs + 1 + 1 + 1 + NumInputs>
+        NumInputs + NumOutputs + 1 + NumInputs + 1>
         unpacked_inputs;
+    // Residual bits from packing of 256-bit long string into 253-bit long field
+    // elements to which are added the public value of size 64 bits
+    const size_t residual_packing_bits =
```
AntoineRondelet: This variable does not represent bits, but rather a `number` of bits or the length of a bit string. Please rename accordingly

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -93,35 +99,99 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
     }
 
     // ============================================================================================ //
+    // Reminder: The primary inputs are either field elements, hash digest or binary arrays.
+    // As a field element may not represent an entire hash digest, we formatted the former in two parts:
+    // a field element and aggregated the remaining bits into (a) common field element(s): the blob(s).
+    // For efficiency we also added in the blob the public values (bit arrays of 64 bits).
```
AntoineRondelet: Please clarify the comments. We need to stick to a single terminology. Here you use `the blob` to refer to the extra bits being packed together, while you refer to this as `residual_packing_bits` in the cpp code.

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -93,35 +99,99 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
     }
 
     // ============================================================================================ //
+    // Reminder: The primary inputs are either field elements, hash digest or binary arrays.
+    // As a field element may not represent an entire hash digest, we formatted the former in two parts:
+    // a field element and aggregated the remaining bits into (a) common field element(s): the blob(s).
```
AntoineRondelet: `common field element(s)`, common bits to what?
Please rephrase to clarify this explanation.

What you do here is basically pack a set of N hash_digest of length > field capacity into a set of N + M field elements. The first 253 (field capacity) bits of the digests are packed into a field element and the remaining bits (hash_digest_length - field_capacity bits) are packed independently into a set of M field elements

AntoineRondelet:
In addition to the review please make sure to rebase this branch on top of develop @rrtoledo

AntoineRondelet:
FYI: You'll need to rebase with develop for the build to pass @rrtoledo

zeth-contracts/contracts/Bytes.sol
```diff
@@ -1,11 +1,11 @@
 pragma solidity ^0.5.0;
 
 library Bytes {
-    function sha256_digest_from_field_elements(uint[] memory input) internal pure returns (bytes32) {
+     function sha256_digest_from_field_elements(uint input, bytes1 bits) internal pure returns (bytes32) {
```
dtebbs: Seems to be indented 5 spaces

src/circuits/joinsplit.tcc
```diff
@@ -118,60 +126,59 @@ public:
...
+            //   NumOutputs + NumInputs + 1[
             // - Index of the "h_sig" field element: {NumOutputs + NumInputs + 1
-            // + 1 + 1}
+            //   }
             // - Index of the "h_iS" field elements: [NumOutputs + NumInputs + 1
-            // + 1 + 1 + 1,  NumOutputs + NumInputs + 1 + 1 + 1 + 1 + NumInputs[
+            //   + 1,  NumOutputs + NumInputs + 1 + NumInputs[
+            // - Index of the "v_pub_in", "v_pub_out", and bits of previous
+            //   varibles not fitting within FieldT::capacity() [NumOutputs +
+            //   NumInputs + 1 + NumInputs, ?]
```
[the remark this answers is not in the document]
rrtoledo: `1 + NumInputs` represents `h_sig` and the `h_is`
`?`  was updated to represent the residual fields

src/circuits/joinsplit.tcc
```diff
@@ -231,44 +247,89 @@ public:
...
                  i++, j++) {
                 unpacked_inputs[i].insert(
                     unpacked_inputs[i].end(),
                     h_is[j]->bits.begin(),
-                    h_is[j]->bits.end());
+                    h_is[j]->bits.begin() + FieldT::capacity());
+            }
+
+            // Initialize the unpacked input corresponding to the variables of
+            // size different to 253 bits (smaller inputs and residual bits).
```
AntoineRondelet: `size different to 253 bits` -> `size different to 253 (FieldT::capacity) bits`

src/circuits/joinsplit.tcc
```diff
@@ -292,8 +353,7 @@ public:
 
             // [SANITY CHECK] Total size of unpacked inputs
             size_t total_size_unpacked_inputs = 0;
-            for (size_t i = 0;
-                 i < NumOutputs + NumInputs + 1 + 1 + 1 + NumInputs;
+            for (size_t i = 0; i < NumInputs + NumOutputs + 1 + NumInputs + 1;
```
AntoineRondelet: Use `nb_field_residual` instead of the final `+ 1` (applicable to other places in this file)

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -93,35 +109,100 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
-            digest_inputs[1] = primary_inputs[i+1];
-            bytes32 current_nullifier = Bytes.sha256_digest_from_field_elements(digest_inputs);
+        uint256 digest_input;
+        bytes1 bits_input;
+        uint index;
+        uint size_extra_bits = digest_length % field_capacity;
+        if (digest_length < field_capacity) {
+            size_extra_bits = 0;
+        }
+        for(uint i = 1; i < 1 + jsIn; i ++) {
```
AntoineRondelet: i ++ -> `i++`

zeth-contracts/contracts/Bytes.sol
```diff
@@ -1,64 +1,55 @@
...
-        // Note, we store the result of 2 ** n in uint8. However, if n is to big (ie: n > 8)
+        // (the first 5 bit of value '0' are just here to fill the space in the byte).
+        // So when we reverse the byte, we have: 0000 0100 --> 0010 0000
+        // But again only 3 bits are meaningful in this case. This time the 5 last bits are padding.
+        // Thus we push the meaningful data to the right. Now we have something like: `0000 0001`.
+        // Note, we store the result of 2 ** n in uint8. However, if n is to big (ie: n > 8).
         // This can overflow. Here this is fine as n is NOT a user input, and does not aim to be changed.
         // Nevertheless we need to keep this in mind. As a consequence we add the "dummy" require below that
         // should fail if we manually change the value of n to be > 8.
+        uint8 aInt = uint8(bits);
```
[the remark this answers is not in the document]
rrtoledo: This notation comes from master: https://github.com/clearmatics/zeth/blob/61a224a43eb7fe98e7a0ad93069fa83fa75b3830/zeth-contracts/contracts/Bytes.sol#L41

I removed this variable and merged the cast and shift as follows
`
uint8 n = 5;
uint8 aInt = uint8(bits); // Converting bytes1 into 8 bit integer
uint8 reversed = uint8(reverse_byte(aInt));
uint8 shifted = reversed / 2 ** n;
bytes1 shifted_byte = bytes1(shifted);
`

`
uint8 reversed = uint8(reverse_byte(uint8(bits)));
uint8 n = 5;
bytes1 shifted_byte = bytes1(reversed) >> n;
`

src/circuits/joinsplit.tcc
```diff
@@ -21,35 +22,43 @@ template<
...
+    // previous variables' bits not containable in a single field element as
+    // well as the public values) (the root is a field element) because we pack
+    // the nullifiers (Inputs of JS = NumInputs), the commitments (Output of JS
+    // = NumOutputs) AND the signature hash h_sig (+1) AND the message
+    // authentication tags h_iS (+ NumInputs) AND the residual field elements
+    // which aggregate the extra bits and public values (+1)
     std::array<
         libsnark::pb_variable_array<FieldT>,
-        NumInputs + NumOutputs + 1 + 1 + 1 + NumInputs>
+        NumInputs + NumOutputs + 1 + NumInputs + 1>
```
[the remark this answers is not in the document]
rrtoledo: I thought it would be cleaner to have 1 pb_variable_array comprising `nb_field_residual` `FieldTs rather than the opposite.

`            // We allocate as many field elements as needed to pack the public
            // values and the hash digests' residual bits
            packed_inputs[NumInputs + NumOutputs + 1 + NumInputs].allocate(
                pb, nb_field_residual, FMT(this->annotation_prefix, " bits"));`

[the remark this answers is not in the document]
AntoineRondelet: I see, thanks

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -93,35 +109,100 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
+    // - v_pub_in [0, size_value[
+    // - v_pub_out [size_value + 1, 2*size_value[
+    // - h_sig remaining bits [2*size_value + 1, 2*size_value + (digest_length-field_capacity)[
+    // - nullifierS remaining bits [2*size_value + (digest_length-field_capacity) + 1, 2*size_value + (1+NumInputs)*(digest_length-field_capacity)[
+    // - commitmentS remaining bits [2*size_value + (1+NumInputs)*(digest_length-field_capacity) + 1, 2*size_value + (1+NumInputs+NumOutputs)*(digest_length-field_capacity)[
+    // - message authentication tagS remaining bits [2*size_value + (1+NumInputs+NumOutputs)*(digest_length-field_capacity) + 1, 2*size_value + (1+2*NumInputs+NumOutputs)*(digest_length-field_capacity)]
     // ============================================================================================ //
 
+    // This function extract the remaining bits of the nullifierS, commitmentS or h_iS from the residual field element(S)
+    function extract_extra_bits(uint start, uint end, uint[] memory primary_inputs) public pure returns (bytes1) {
```
AntoineRondelet: Please provide tests for the newly introduced functions

zeth-contracts/contracts/Bytes.sol
```diff
@@ -145,8 +134,18 @@ library Bytes {
     // Output: 0001 0000 (binary) -> 16 (decimal)
     function reverse_byte(uint a) internal pure returns (uint) {
         uint c = 0xf070b030d0509010e060a020c0408000;
-
-        return (( c >> ((a & 0xF)*8)) & 0xF0)   +
+        return (( c >> ((a & 0xF)*8)) & 0xF0) +
             (( c >> (((a >> 4)&0xF)*8) + 4) & 0xF);
     }
+
+    function swap_bit_order(bytes32 input) internal pure returns (bytes32) {
```
AntoineRondelet: Please add a test for this function

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -93,35 +109,100 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
         // 2. We re-assemble the nullifiers (JSInputs)
-        for(uint i = 1; i < 1 + 2*jsIn; i += 2) {
-            digest_inputs[0] = primary_inputs[i];
-            digest_inputs[1] = primary_inputs[i+1];
-            bytes32 current_nullifier = Bytes.sha256_digest_from_field_elements(digest_inputs);
+        uint256 digest_input;
+        bytes1 bits_input;
+        uint index;
+        uint size_extra_bits = digest_length % field_capacity;
+        if (digest_length < field_capacity) {
```
AntoineRondelet: Should be `<=` I think

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -130,73 +211,40 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
-
-
     function assemble_commitments_and_append_to_state(uint[] memory primary_inputs) internal {
         // We re-assemble the commitments (JSOutputs)
-        uint256[] memory digest_inputs = new uint[](2);
-        for(uint i = 1 + 2 * jsIn ; i < 1 + 2*(jsIn + jsOut); i += 2) {
-            digest_inputs[0] = primary_inputs[i]; // See the way the inputs are ordered in the extended proof
-            digest_inputs[1] = primary_inputs[i+1];
-            bytes32 current_commitment = Bytes.sha256_digest_from_field_elements(digest_inputs);
+        uint256 digest_input;
```
AntoineRondelet: Is this variable really useful, can't you just replace it by `primary_inputs[i]` throughout the function?

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -93,35 +109,100 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
+        // as such we compute the indices of the first and last field elements
+        // they are located at
+        uint index_start_bytes32 = 1 + 1 + 2*jsIn + jsOut + (offset+start)/field_capacity;
+        uint index_end_bytes32 = 1 + 1 + 2*jsIn + jsOut + (offset+end)/field_capacity;
+
+        // We compute the padding length in the field elements containing the remaining bits
+        uint padding_start = digest_length - field_capacity;
+        uint padding_end = padding_start;
+        if (primary_inputs.length-1 == index_end_bytes32) {
+            padding_end = digest_length - ( (offset + (digest_length - field_capacity)*(2*jsIn + jsOut)) % field_capacity);
```
AntoineRondelet: You `padding` variables are confusing. You subtract the offset, then add it again in `uint index_start_byte = (padding_start+offset+start)/8;`

Please simplify things that can be, or at very least add comments to document your code.

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -52,13 +52,29 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
-    uint constant nbInputs = 1 + 2 * (jsIn + jsOut) + 1 + 1 + 2 * (1 + jsIn);
+    // Constants regarding the hash digest length, the prime number used and its associated length in bits and the max values (v_in and v_out)
+    uint constant digest_length = 256;
+    // uint r = 21888242871839275222246405745257275088548364400416034343698204186575808495617;
+    // field_capacity = ceil ( log_2(r) ) - 1
+    uint constant field_capacity = 253;
+    uint constant size_value = 64;
+
+    // Number of residual bits from packing of 256-bit long string into 253-bit long field elements to which are added the public value of size 64 bits
+    uint constant length_bit_residual = 2 * 64 + (digest_length - field_capacity) * (1 + 2 * jsIn + jsOut);
```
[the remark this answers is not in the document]
rrtoledo: I missed it thks!

zeth-contracts/contracts/BaseMixer.sol
rrtoledo: Sure. 

We do not have any test files for the functions in BaseMixer.sol so far. I'll create a specific one for this.

zeth-contracts/contracts/BaseMixer.sol
rrtoledo: Yes. Corrected.

zeth-contracts/contracts/BaseMixer.sol
rrtoledo: Yes. I will add some comments.

The offset represents the 128 bits of the public values and 3 bits for the h_sig, while the padding represent just padding. What I am trying to do here is say that the padding is 3 bit if the field element is full, else 256 - the remaining bits.

zeth-contracts/contracts/BaseMixer_tests.sol
```diff
@@ -0,0 +1,52 @@
...
+
+contract BaseMixer_tests {
+	using Bytes for *;
+
+	constructor() public {
+        // Nothing
+    }
+
+    function test_extract_extra_bits() public returns (bool) {
+        // test_bytes = primary inputs
```
AntoineRondelet: Please make sure to clean up the indentation in the files of the repo

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -130,73 +216,38 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
+        uint index;
+        bytes1 bits_input;
+        uint size_extra_bits = digest_length % field_capacity;
+        if (digest_length < field_capacity) {
+            size_extra_bits = 0;
+        }
+        for(uint i = 1 + jsIn ; i < 1 + jsIn + jsOut; i ++) {
+            // See the way the inputs are ordered in the extended proof
+            index = (digest_length-field_capacity)*(i-1);
+            bits_input = extract_extra_bits(index, index+(size_extra_bits-1), primary_inputs);
```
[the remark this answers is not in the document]
rrtoledo: I used the modulo (`uint size_extra_bits = digest_length % field_capacity;`) as a generalisation (say we have `digest_length = 2*field_capacity + a` for some `a`).

The `-1` is not to extract the last position asked. I can see how this leads to a problem. A solution would be to change the function definitions so that the second argument is the number of bits we want to extract and not the position of the last bit we want to extract.

I wrote `size_extra_bits = 0` to have an edge case where we do not need to extract any bits. The `-1` however brings an error (we'll have the start position > end position, and potentially the end position equal to `-1` resulting in an overflow). The previous solution should resolve it.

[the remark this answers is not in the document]
AntoineRondelet: > I did not use modulo consistently however which may lead to problems (if a simple substraction is not enough).

Yes

> A solution would be to change the function definitions so that the second argument is the number of bits we want to extract and not the position of the last bit we want to extract

I like this solution better yes

zeth-contracts/contracts/BaseMixer_tests.sol
rrtoledo: Done

zeth-contracts/contracts/BaseMixer.sol
rrtoledo: Corrected

zeth-contracts/contracts/BaseMixer.sol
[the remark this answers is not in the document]
rrtoledo: Fixed.

src/circuits/joinsplit.tcc
rrtoledo: Done

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -93,35 +111,97 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
+        uint offset = 2 * size_value + length_extra_bits;
+
+        // We then compute the position of the last bit to retrieve
+        uint end = start + length - 1;
+
+        // The remaining bits could be written over two field elements
+        // as such we compute the indices of the first and last field elements
+        // they are located at
+        uint index_start_bytes32 = 1 + 1 + 2*jsIn + jsOut + (offset+start)/field_capacity;
+        uint index_end_bytes32 = index_start_bytes32 + (length-1)/field_capacity;
```
AntoineRondelet: Since `length` is constrainted to be <16, `(length-1)/field_capacity` will always be 0, and so `index_end_bytes32 = index_start_bytes32`, right?

AntoineRondelet: You should do `uint index_end_bytes32 = 1 + 1 + 2*jsIn + jsOut + (offset+start+length-1)/field_capacity;` if I get your intention correctly. In fact, `(offset+start)/field_capacity` will be div_floored, and `(length-1)/field_capacity` = 0

rrtoledo: Yes you are right.

I tried to do an optimization and the tests passed only because of our particular case. I'll correct that right away.

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -135,52 +137,48 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
 
         // We now recombine the two bytes
         // We multiply start_byte by 256 = 2**8.
-        res = start_byte*(2**8)+end_byte;
+        uint16 res = start_byte*(2**8) + end_byte;
         // And remove any unneeded bits
-        uint index_start_bit = (offset + padding_start + start) % 8;
-        res = res * uint16(2**index_start_bit);
-        res = res / uint16(2**(16 - (end - start + 1)));
+        res = res * uint16(2**(padded_start % 8));
```
AntoineRondelet: ` uint padded_start = start + offset + ((primary_inputs.length-1 == index_start_bytes32) ? potential_padding : length_extra_bits);`

and now you do:
> res = res * uint16(2**(padded_start % 8));

If this supposed to be equivalent to what you did in your previous commit:
```
uint index_start_bit = (offset + padding_start + start) % 8;
```
Because it does not seem to be.

zeth-contracts/contracts/BaseMixer.sol
AntoineRondelet: Let's have something working first, and we'll think about optimizations in a second time. For now, we should focus on having a **documented and clear code** that works

AntoineRondelet: https://github.com/clearmatics/zeth/pull/104

zeth-contracts/contracts/BaseMixer.sol
rrtoledo: It should be equivalent:
We have that
`padding_start = digest_length - ( (offset + (digest_length - field_capacity)*(2*jsIn + jsOut)) % field_capacity`  if `primary_inputs.length-1 == index_start_bytes32`
else 
`uint potential_padding = digest_length - ((offset + length_extra_bits*(2*jsIn + jsOut)) % field_capacity);` is the same as before and by definition `length_extra_bits` equals `digest_length - field_capacity`

The main difference is than now if we ask for a length of 0 we directly leave.

On the other side we have:
`uint padded_start = start + offset + ((primary_inputs.length-1 == index_start_bytes32) ? potential_padding : length_extra_bits);`
So "padding start" should be: `((primary_inputs.length-1 == index_start_bytes32) ? potential_padding : length_extra_bits);`
`potential_padding`

zeth-contracts/contracts/BaseMixer_tests.sol
```diff
@@ -0,0 +1,51 @@
+pragma solidity ^0.5.0;
+
+import './BaseMixer.sol';
+
+contract BaseMixer_tests {
```
AntoineRondelet: Throughout this PR, and https://github.com/clearmatics/zeth/pull/104 we bumped into several edge cases that should be fixed now:
1. The `end` pointer (now removed) was potentially smaller than `start`
2.  The `require` on `length` was not properly set

Thus, for the tests to be meaningful, we need to think about these few edge cases and trigger them in the tests to make sure that the flaws patched in the PR have properly been patched.
Hence, please add test cases to hit the `require`, to pass a length == 0 (hit the return), and other potential scenarios you feel are relevant to test the fixes discussed in this PR

rrtoledo: I totally agree.
I intend to add a few more tests. 

I want to cover the following cases:
- length of 0
- length within the `require`
- length above the `require`
And for length within the `require`:
- bits spread over two field elements
- bits spread over one field element but two bytes
- bits spread over one byte of a field element.

I also intend to add a `require` on the size of  the function input `primary_inputs` (we might try to retrieve non existing elements) and add some tests on it.

rrtoledo: I forgot: we also need to constraint `start` and `start + length - 1`.

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -58,19 +58,43 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
+    uint nbInputs = 1 + jsIn + jsOut + 1 + jsIn + nb_field_residual;
 
     // Contract variable that indicates the address of the token contract
     // If token = address(0) then the mixer works with ether
     address public token;
 
     // The unit used for public values (ether in and out), in Wei. Must match
     // the python wrappers. Use Szabos (10^12 Wei).
-    uint64 constant public_unit_value_wei = 1 szabo;
+    uint constant public_unit_value_wei = 1 szabo;
```
dtebbs: Was there some reason to change the type here?

rrtoledo: No reason! 
It must be due to a rebase and I missed it, changing it now.

rrtoledo: @dtebbs corrected.
I had to update  `vpub_in_zeth_units;` to uint however to prevent an integer overflow (we discussed about this in november on slack).

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -103,15 +127,134 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
+        uint padding = packing_residue_length;
+        // Update the padding if the public values and remainder bits fit in one field element
+        if (2*size_value+padding*(1+2*jsIn+jsOut) < field_capacity) {
+            padding = digest_length - (2*size_value + packing_residue_length*(1+2*jsIn+jsOut));
+        }
+
+        // We retrieve the public value in, remove any extra bits (due to the padding) and inverse the bit order
+        bytes32 vpub_bytes = (bytes32(primary_inputs[1 + 1 + 2*jsIn + jsOut]) << padding) >> (digest_length-size_value);
+        vpub_bytes = Bytes.swap_bit_order(vpub_bytes) >> (digest_length-size_value);
+        vpub_in = Bytes.get_int64_from_bytes8(Bytes.int256ToBytes8(uint(vpub_bytes)));
```
dtebbs: The shift left and then back down seems to be unnecessary.  Can't you just do:
```
        bytes32 vpub_bytes = (bytes32(primary_inputs[1 + 1 + 2*jsIn + jsOut]) << padding);
        vpub_bytes = Bytes.swap_bit_order(vpub_bytes);
        vpub_in = Bytes.get_int64_from_bytes8(Bytes.int256ToBytes8(uint(vpub_bytes)));
```

dtebbs: BTW, we can leave optimizations for #138, so no need to change this now if you find it clearer.

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -103,15 +127,134 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
     // ============================================================================================ //
     // Reminder: Remember that the primary inputs are ordered as follows:
     // We make sure to have the primary inputs ordered as follow:
-    // [Root, NullifierS, CommitmentS, value_pub_in, value_pub_out]
+    // [Root, NullifierS, CommitmentS, h_sig, h_iS, Residual Field Element(S)]
```
dtebbs: Have you considered moving h_sig and everything beyond to before the `NullifierS`?  In this way, the code to reassemble these values shoudl be much simpler (no matter how many inputs / outpus, we can be sure that we can fit the public values and residual parts of elements into a single field element, so we avoid a lot `if (2*size_value+padding*(1+2*jsIn+jsOut) < field_capacity) ....` conditions.

Again, not super important at this stage, but could simplify some of the code.

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -103,15 +127,134 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
+
+    // This function is used to reassemble hsig given the the primary_inputs
+    // To do so, we extract the remaining bits of hsig from the residual field element(S) and combine them with the hsig field element
+    function assemble_hsig(uint[] memory primary_inputs) public view returns (bytes32 hsig){
+
+        // We know hsig residual bits correspond to the 128th to 130st bits of the first residual field element after padding.
+        // We start by computing the padding.
+        uint padding = packing_residue_length;
+        // Update the padding if the public values and remainder bits fit in one field element
+        if (2*size_value+padding*(1+2*jsIn+jsOut) < field_capacity) {
```
dtebbs: If thsi condition does not hold, do you not need to also adjust the index `1 + 1 + 2*jsIn + jsOut` into `primary_inputs` that the residuals get taken from?  (Similarly in other `assemble_*` functions).  Or, perhaps it's safer to just `require` that the residuals + public values don't overflow a single field element.

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -58,11 +58,35 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
+
+    // Constants regarding the hash digest length, the prime number used and its associated length in bits and the max values (v_in and v_out)
+    // uint r = 21888242871839275222246405745257275088548364400416034343698204186575808495617;
+    // field_capacity = ceil ( log_2(r) ) - 1
+    uint constant digest_length = 256;
+    uint constant field_capacity = 253;
+
+    // Variable representing the number of "residual" bits we can expect from converting a hash digest into a field element
+    // see primary input `residual_bits` in Reminder below
+    uint packing_residue_length = digest_length > field_capacity ? digest_length % field_capacity : 0;
```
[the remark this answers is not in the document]
rrtoledo: it works !
Changed, thanks. :)

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -58,11 +58,35 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
+
+    // Total number of residual bits from packing of 256-bit long string into 253-bit long field elements
+    // to which are added the public value of size 64 bits
+    uint length_bit_residual = 2 * size_value + packing_residue_length * (1 + 2 * jsIn + jsOut);
```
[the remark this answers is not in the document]
rrtoledo: Changed, thanks. :)

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -58,11 +58,35 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
+    uint nb_field_residual = (length_bit_residual + field_capacity - 1) / field_capacity;
+
+    // The number of public inputs is:
+    // - 1 (the root)
+    // - jsIn (the nullifiers)
+    // - jsOut (the commitments)
+    // - 1 (hsig)
+    // - JsIn (the message auth. tags)
+    // - nb_field_residual (the residual bits not fitting in a single field element and the in and out public values)
+    uint nbInputs = 1 + jsIn + jsOut + 1 + jsIn + nb_field_residual;
```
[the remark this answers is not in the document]
rrtoledo: Changed, thanks. :)

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -58,11 +58,35 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
     // The number or public inputs is: 1 (the root) + 2 for each digest (nullifiers, commitments) + (1 + 1) (in and out public values) field elements
     uint constant jsIn = 2; // Nb of nullifiers
-    uint constant jsOut = 2; // Nb of commitments/ciphertexts
+    uint constant jsOut = 2; // Nb of commitments/cipher-texts
 
-    // We have 2 field elements for each digest (nullifierS (jsIn), commitmentS (jsOut), h_iS (jsIn) and h_sig)
-    // The root, v_pub_in and v_pub_out are all represented by one field element, so we have 1 + 1 + 1 extra public values
-    uint constant nbInputs = 1 + 2 * (jsIn + jsOut) + 1 + 1 + 2 * (1 + jsIn);
+    // Size of the public values in bits
+    uint constant size_value = 64;
```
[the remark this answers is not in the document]
rrtoledo: Sure. 
Changed to `public_value_length`, thanks. :)

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -103,15 +127,134 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
     // - Index of the "CommitmentS" field elements: [NumInputs + 1, NumOutputs + NumInputs + 1[
-    // - Index of the "v_pub_in" field element: {NumOutputs + NumInputs + 1}
-    // - Index of the "v_pub_out" field element: {NumOutputs + NumInputs + 1 + 1}
+    // - Index of the "h_sig" field element: {NumOutputs + NumInputs + 1}
+    // - Index of the "Message Authentication TagS" (h_i) field elements: [NumOutputs + NumInputs + 1 + 1, NumOutputs + NumInputs + 1 + NumOuputs [
+    // - Index of the "Residual Field Element(s)" field elements: [NumOutputs + NumInputs + 1 + NumOuputs + 1 , NumOutputs + NumInputs + 1 + NumOuputs + 1 + nb_field_residual]
+    //
+    // The Residual field elements are structured as follows:
+    // - v_pub_in [0, size_value[
+    // - v_pub_out [size_value + 1, 2*size_value[
```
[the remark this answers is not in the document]
rrtoledo: My bad.
All updated. Thanks !

zeth-contracts/contracts/BaseMixer.sol
rrtoledo: No problem, this works well. I might just add a comment that `int256ToBytes8` extracts the last 8 bytes.

N.B. if we want to reverse the bit order, I guess we can just do:
`
        uint residual_hash_size = packing_residue_length*(1+2*jsIn+jsOut);
        bytes32 vpub_bytes = (bytes32(primary_inputs[1 + 1 + 2*jsIn + jsOut]) >> residual_hash_size + public_value_length);
        vpub_in = Bytes.get_int64_from_bytes8(Bytes.int256ToBytes8(uint(vpub_bytes)));

        vpub_bytes = bytes32(primary_inputs[1 + 1 + 2*jsIn + jsOut]) >> (residual_hash_size);
        vpub_out = Bytes.get_int64_from_bytes8(Bytes.int256ToBytes8(uint(vpub_bytes)));
`

zeth-contracts/contracts/BaseMixer.sol
rrtoledo: Yes, I assumed that the condition holds.
(It was quite expensive to consider the case where we have more than one field element as residual bits since we have to handle 2 bytes at a time in the worst case scenario).

I am not sure where to add the following require however:
```
    require(
        2*public_value_length+packing_residue_length*(1+2*jsIn+jsOut) < field_capacity,
        "too many input and output notes considered"
    );
```

zeth-contracts/contracts/BaseMixer.sol
rrtoledo: I assumed we had one field element (c.f. next comment).

zeth-contracts/contracts/BaseMixer.sol
rrtoledo: After talking, we decided that the constructor was the best place to write this require.

pyClient/test/test_contract_base_mixer.py
```diff
@@ -0,0 +1,314 @@
...
+def test_extract(mixer_instance: Any) -> int:
+    # This test checks that we extract the right values
+    # for {sn}, {cm} and {h} regardless of whether
+    # the bits are located in the same byte
+    # or two different ones (c.f. residual_bits above).
+    print("--- testing ", "test_extract")
+
+    start = 0
+    for i in range(2*js_in+js_out):
+        res = mixer_instance.functions.extract_extra_bits(
```
AntoineRondelet: Hum... can you point me to the line where this function is defined in the contract? I don't find the definition of this function

pyClient/test/test_contract_base_mixer.py
```diff
@@ -0,0 +1,314 @@
...
+import zeth.joinsplit
+
+from typing import Any
+
+# Mixer variables
+size_value = 64  # pylint: disable=no-member,invalid-name
+digest_length = 256  # pylint: disable=no-member,invalid-name
+field_capacity = 253  # pylint: disable=no-member,invalid-name
+js_in = 2  # pylint: disable=no-member,invalid-name
+js_out = 2  # pylint: disable=no-member,invalid-name
```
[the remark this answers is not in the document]
AntoineRondelet: Moreover we should write constants in upper case

pyClient/test/test_contract_base_mixer.py
```diff
@@ -0,0 +1,314 @@
...
+    return 1
+
+
+def test_assemble_commitments(mixer_instance: Any) -> int:
+    # Test retrieving commitments
+    print("--- testing ", "test_assemble_commitments")
+    for i in range(js_out):
+        res = mixer_instance.functions.assemble_commitment(i, inputs).call()
+        val = int.from_bytes(res, byteorder="big")
+        expected_val = 2**(256-1-1) + int(("{0:03b}".format(2+i))[::-1], 2)
```
AntoineRondelet: why `2**(256-1-1)`, and not `2**(256-1)`? the expected value should be a digest of 256 bits, so the leading bit should correspond to `2**(256-1)`

AntoineRondelet: Hum... ok I think I understand now... why not just define the `expected_var` from the `inputs` you define at the beginning of the document instead of doing it from the loop index. It's just confusing.

pyClient/test/test_contract_base_mixer.py
```diff
@@ -0,0 +1,314 @@
...
+    start = 0
+    for i in range(2*js_in+js_out):
+        res = mixer_instance.functions.extract_extra_bits(
+            start,
+            packing_residue_length,
+            inputs
+            ).call()
+        if int.from_bytes(res, byteorder="big") != i:
+            print("ERROR: extracted wrong value")
+            return 0
```
AntoineRondelet: Why return `0`? 0 indicates a success, but you have an error message here. It should be =/= 0

pyClient/test/test_contract_base_mixer.py
```diff
@@ -0,0 +1,314 @@
...
+    for i in range(js_in):
+        res = mixer_instance.functions.assemble_nullifier(i, inputs).call()
+        val = int.from_bytes(res, byteorder="big")
+        expected_val = 2**(256-1) + int(("{0:03b}".format(i))[::-1], 2)
+        if val != expected_val:
+            print("ERROR: extracted wrong nullifier")
+            print("expected:", expected_val, i)
+            print("got:", val, i)
+            return 0
+    return 1
```
AntoineRondelet: should be return 0 to illustrate a successful test

src/circuits/joinsplit.tcc
```diff
@@ -25,35 +25,43 @@ template<
...
-    // val_pub_out) (the root is a field element) `NumInputs + NumOutputs`
-    // because we pack the nullifiers (Inputs of JS = NumInputs), the
-    // commitments (Output of JS = NumOutputs) AND the v_pub_out taken out of
-    // the mix (+1) AND the public value v_pub_in that is put into the mix (+1)
-    // AND the signature hash h_sig (+1) AND the message authentication tags
-    // h_iS (+NumInputs)
+    // Number of residual bits from packing of hash digests into smaller
+    // field elements to which are added the public value of size 64 bits
+    const size_t length_bit_residual =
+        2 * 64 + (HashT::get_digest_len() - FieldT::capacity()) *
```
AntoineRondelet: Define 64 (public values bit length) as a constant and use it here

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -56,13 +56,41 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
     // JoinSplit description, gives the number of inputs (nullifiers) and outputs (commitments/ciphertexts) to receive and process
     // IMPORTANT NOTE: We need to employ the same JS configuration than the one used in the cpp prover
     // Here we use 2 inputs and 2 outputs (it is a 2-2 JS)
-    // The number or public inputs is: 1 (the root) + 2 for each digest (nullifiers, commitments) + (1 + 1) (in and out public values) field elements
     uint constant jsIn = 2; // Nb of nullifiers
-    uint constant jsOut = 2; // Nb of commitments/ciphertexts
+    uint constant jsOut = 2; // Nb of commitments/cipher-texts
```
AntoineRondelet: ciphertext (https://en.wikipedia.org/wiki/Ciphertext)

rrtoledo: Fixed

src/circuits/joinsplit.tcc
rrtoledo: Changed (and other 64s in the file) to `ZETH_V_SIZE` from `zeth.h`

pyClient/test/test_contract_base_mixer.py
rrtoledo: I am used to returning 1 if the test passes, 0 otherwise.
Will update to "0=pass" convention.

rrtoledo: Fixed

pyClient/test/test_contract_base_mixer.py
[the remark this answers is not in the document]
rrtoledo: Updated

pyClient/test/test_contract_base_mixer.py
rrtoledo: I forgot to update the test contract. This function does not exist anymore, we now have helper functions for each kind of primary inputs.

The tests have been updated.

pyClient/test/test_contract_base_mixer.py
rrtoledo: Updated with `inputs[ ]`

pyClient/test/test_contract_base_mixer.py
rrtoledo: Fixed

src/util.cpp
```diff
@@ -120,22 +120,28 @@ std::vector<bool> hex_to_binary_vector(std::string hex_str)
...
 {
-    return get_bits256_from_vector(hex_digest_to_binary_vector(str));
+    if (str.length() != 64) {
+        throw std::length_error(
+            "Invalid string length for the given hex digest (should be "
+            "64)");
+    }
+
+    return get_bits256_from_vector(hex_to_binary_vector(str));
 }
```
AntoineRondelet: Just realized that in https://github.com/clearmatics/zeth/blob/packing/src/test/note_test.cpp we do: 
```
get_bits256_from_vector(hex_digest_to_binary_vector("FF0000000000000000000000000000000000000000000000000000000000000F"));
```

and
```
bits64 value_bits64 = get_bits64_from_vector(hex_to_binary_vector("2F0000000000000F"));
```

Let's just replace `get_bits64_from_vector(hex_to_binary_vector(` by `hex_value_to_bits64`
and `get_bits256_from_vector(hex_digest_to_binary_vector(` by `hex_digest_to_bits256`

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -126,47 +256,42 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
     function assemble_commitments_and_append_to_state(uint[] memory primary_inputs) internal {
         // We re-assemble the commitments (JSOutputs)
-        uint256[] memory digest_inputs = new uint[](2);
-        for(uint i = 1 + 2 * jsIn ; i < 1 + 2*(jsIn + jsOut); i += 2) {
-            digest_inputs[0] = primary_inputs[i]; // See the way the inputs are ordered in the extended proof
-            digest_inputs[1] = primary_inputs[i+1];
-            bytes32 current_commitment = Bytes.sha256_digest_from_field_elements(digest_inputs);
-            uint commitment_address = insert(current_commitment);
-            emit LogCommitment(commitment_address, current_commitment);
+        for(uint i ; i < jsOut; i ++) {
```
AntoineRondelet: `for(uint i ; i < jsOut; i ++)` -> `for (uint i; i < jsOut; i++)`

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -98,20 +126,122 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
+    // - commitmentS remaining bits [2*public_value_length + (1+NumInputs)*(digest_length-field_capacity), 2*public_value_length + (1+NumInputs+NumOutputs)*(digest_length-field_capacity)[
+    // - message authentication tagS remaining bits [2*public_value_length + (1+NumInputs+NumOutputs)*(digest_length-field_capacity), 2*public_value_length + (1+2*NumInputs+NumOutputs)*(digest_length-field_capacity)]
     // ============================================================================================ //
 
+    // This function is used to extract the public values (vpub_in and vpub_out) from the residual field element(S)
+    function assemble_public_values(uint[] memory primary_inputs) public pure returns (uint64 vpub_in, uint64 vpub_out){
+        // We know vpub_in corresponds to the first 64 bits of the first residual field element after padding.
+        // We retrieve the public value in, remove any extra bits (due to the padding) and inverse the bit order
+        // N.B. Bytes.int256ToBytes8(x) extract the last 8 bytes of x
```
AntoineRondelet: `int256ToBytes8` some function names are not consistent in `Bytes.sol`. This dates back to a long time ago and is not strictly relevant to this PR. I'm just flagging it as a quick thing to fix, in the future, to improve code consistency

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -56,13 +56,41 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
+    // uint r = 21888242871839275222246405745257275088548364400416034343698204186575808495617;
+    // field_capacity = floor( log_2(r) )
+    uint constant digest_length = 256;
+    uint constant field_capacity = 253;
+
+    // Variable representing the number of "residual" bits we can expect from converting a hash digest into a field element
+    // see primary input `residual_bits` in Reminder below
+    uint constant packing_residue_length = digest_length - field_capacity;
+
+    // Number of hash digests int he primary inputs
```
AntoineRondelet: `int he` -> `in the`

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -98,20 +126,122 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
 
     // ============================================================================================ //
     // Reminder: Remember that the primary inputs are ordered as follows:
     // We make sure to have the primary inputs ordered as follow:
-    // [Root, NullifierS, CommitmentS, value_pub_in, value_pub_out]
+    // [Root, NullifierS, CommitmentS, h_sig, h_iS, Residual Field Element(S)]
     // ie, below is the index mapping of the primary input elements on the protoboard:
     // - Index of the "Root" field elements: {0}
     // - Index of the "NullifierS" field elements: [1, NumInputs + 1[
     // - Index of the "CommitmentS" field elements: [NumInputs + 1, NumOutputs + NumInputs + 1[
```
AntoineRondelet: `// - Index of the "CommitmentS" field elements: [1+ NumInputs, 1 + NumInputs + NumOutputs[`

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -98,20 +126,122 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
     }
 
     // ============================================================================================ //
     // Reminder: Remember that the primary inputs are ordered as follows:
     // We make sure to have the primary inputs ordered as follow:
-    // [Root, NullifierS, CommitmentS, value_pub_in, value_pub_out]
+    // [Root, NullifierS, CommitmentS, h_sig, h_iS, Residual Field Element(S)]
     // ie, below is the index mapping of the primary input elements on the protoboard:
     // - Index of the "Root" field elements: {0}
     // - Index of the "NullifierS" field elements: [1, NumInputs + 1[
```
AntoineRondelet: `// - Index of the "NullifierS" field elements: [1, NumInputs + 1[` -> `// - Index of the "NullifierS" field elements: [1, 1 + NumInputs[`

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -98,20 +126,122 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
     // - Index of the "CommitmentS" field elements: [NumInputs + 1, NumOutputs + NumInputs + 1[
-    // - Index of the "v_pub_in" field element: {NumOutputs + NumInputs + 1}
-    // - Index of the "v_pub_out" field element: {NumOutputs + NumInputs + 1 + 1}
+    // - Index of the "h_sig" field element: {NumOutputs + NumInputs + 1}
```
AntoineRondelet: `{1 + NumInputs + NumOutputs}`

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -98,20 +126,122 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
+    // - Index of the "Message Authentication TagS" (h_i) field elements: [NumOutputs + NumInputs + 1 + 1, NumOutputs + NumInputs + 1 + NumOuputs [
```
AntoineRondelet: `Index of the "Message Authentication TagS" (h_i) field elements: [1 + NumInputs + NumOutputs + 1, 1 + NumInputs + NumOutputs + 1 + NumInputs[`

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -98,20 +126,122 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
+    // - Index of the "Residual Field Element(s)" field elements: [NumOutputs + NumInputs + 1 + NumOuputs + 1 , NumOutputs + NumInputs + 1 + NumOuputs + 1 + nb_field_residual]
```
AntoineRondelet: `// - Index of the "Residual Field Element(s)" field elements: [1 + NumInputs + NumOutputs + 1 + NumInputs, 1 + NumInputs + NumOutputs + 1 + NumInputs + nb_field_residual]`

zeth-contracts/contracts/BaseMixer.sol
AntoineRondelet: There are `NumInputs` htags, not `NumOutput`

src/util.cpp
AntoineRondelet: Note: This can be fixed in another PR as it is not strictly related to this one

zeth-contracts/contracts/BaseMixer.sol
AntoineRondelet: Note: This can be fixed in another PR as it is not strictly related to this one

pyClient/test/test_contract_base_mixer.py
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env python3
+
+# Copyright (c) 2015-2019 Clearmatics Technologies Ltd
+#
+# SPDX-License-Identifier: LGPL-3.0+
```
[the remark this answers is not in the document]
rrtoledo: Fixed.

zeth-contracts/contracts/BaseMixer.sol
rrtoledo: Fixed.

zeth-contracts/contracts/BaseMixer.sol
rrtoledo: Function name changed to `extract_last_8_bytes`

zeth-contracts/contracts/BaseMixer.sol
rrtoledo: Fixed

zeth-contracts/contracts/BaseMixer.sol
rrtoledo: Fixed

zeth-contracts/contracts/BaseMixer.sol
rrtoledo: Fixed

zeth-contracts/contracts/BaseMixer.sol
rrtoledo: Fixed

zeth-contracts/contracts/BaseMixer.sol
rrtoledo: Fixed

zeth-contracts/contracts/BaseMixer.sol
rrtoledo: Fixed.
I think it should be `[ ... + nb_field_residual[`

src/util.cpp
rrtoledo: Fixed in PR #151

pyClient/test/test_contract_base_mixer.py
AntoineRondelet: Alright, that's what I thought

pyClient/test/test_contract_base_mixer.py
```diff
@@ -0,0 +1,204 @@
...
+            assemble_commitment(i, PRIMARY_INPUTS).call()
+        val = int.from_bytes(res, byteorder="big")
+        # We need to recompute the expected value
+        # To do so, we load the variable's first FIELD_CAPACITY bits from
+        # `PRIMARY_INPUTS` and remove the padding
+        # (of size DIGEST_LENGTH-FIELD_CAPACITY)
+        # Before adding the value defined in `residual bits`
+        expected_val = PRIMARY_INPUTS[1+JS_INPUTS+i] *\
+            2**(DIGEST_LENGTH-FIELD_CAPACITY) +\
+            RESIDUAL_BITS[3+JS_INPUTS+i]
```
AntoineRondelet: I found these to be confusing, which is problematic since it should be effortless to reason about such a simple test.

Let's modify the test and the comments by:
1. Providing the inputs before packing (the expected values)
2. Writing the inputs after packing (basically as you did)
3. Use the values in 1. as expected values instead of recomposing the expected values from either the loop index or the packed values

rrtoledo: Ok!

AntoineRondelet: `PRIMARY_INPUTS` should be updated to `PACKED_PRIMARY_INPUTS`, and we need to detail from which unpacked primary_inputs we obtained them to see if we recover the same values at the end

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -126,47 +255,42 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
     function assemble_commitments_and_append_to_state(uint[] memory primary_inputs) internal {
         // We re-assemble the commitments (JSOutputs)
-        uint256[] memory digest_inputs = new uint[](2);
-        for(uint i = 1 + 2 * jsIn ; i < 1 + 2*(jsIn + jsOut); i += 2) {
-            digest_inputs[0] = primary_inputs[i]; // See the way the inputs are ordered in the extended proof
-            digest_inputs[1] = primary_inputs[i+1];
-            bytes32 current_commitment = Bytes.sha256_digest_from_field_elements(digest_inputs);
-            uint commitment_address = insert(current_commitment);
-            emit LogCommitment(commitment_address, current_commitment);
+        for(uint i ; i < jsOut; i++) {
```
AntoineRondelet: https://github.com/clearmatics/zeth/pull/80#discussion_r370655845

Remove space between `uint i` and `;`

zeth-contracts/contracts/BaseMixer.sol
AntoineRondelet: yes `[` indeed

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -98,20 +126,121 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
     // ie, below is the index mapping of the primary input elements on the protoboard:
     // - Index of the "Root" field elements: {0}
-    // - Index of the "NullifierS" field elements: [1, NumInputs + 1[
-    // - Index of the "CommitmentS" field elements: [NumInputs + 1, NumOutputs + NumInputs + 1[
-    // - Index of the "v_pub_in" field element: {NumOutputs + NumInputs + 1}
-    // - Index of the "v_pub_out" field element: {NumOutputs + NumInputs + 1 + 1}
+    // - Index of the "NullifierS" field elements: [1, 1 + NumInputs[
+    // - Index of the "CommitmentS" field elements: [1 + NumInputs, 1 + NumInputs + NumOutputs[
+    // - Index of the "h_sig" field element: {1 + NumInputs + NumOutputs}
+    // - Index of the "Message Authentication TagS" (h_i) field elements: [1 + NumInputs + NumOutputs + 1, 1 + NumInputs + NumOuputs + 1 + NumInputs [
```
AntoineRondelet: `NumInputs [` -> `NumInputs[`

pyClient/test/test_contract_base_mixer.py
rrtoledo: > and we need to detail from which unpacked primary_inputs we obtained them 

I am not sure if you want to differentiate the unpacked primary from one another (as follows) or if you are saying something else.
```
UNPACKED_SN = [
    8,
    9
]
```

AntoineRondelet: You can write them all in the same array, with a comment next to it to specify which input it is, e.g. `8, # sn_0`.


Btw, we should stick to `nullifier` instead of `serial number`

rrtoledo: Fixed.

zeth-contracts/contracts/BaseMixer.sol
rrtoledo: Fixed.

zeth-contracts/contracts/BaseMixer.sol
rrtoledo: Fixed

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -98,20 +126,121 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
+        require(
+            field_capacity >= sn_bit_index + packing_residue_length,
+            "nullifier written in different residual bit f.e."
+        );
+
+        // We retrieve nf's residual bits and remove any extra bits (due to the padding)
+        bytes32 sn_bytes = (bytes32(primary_inputs[1 + nb_hash_digests]) << padding_size + sn_bit_index) >> field_capacity;
+        bytes1 bits_input = Bytes.get_last_byte(sn_bytes);
+        // We reassemble the residual bits with the field element
+        sn = Bytes.sha256_digest_from_field_elements(primary_inputs[nullifier_index] << (digest_length - field_capacity), bits_input);
```
AntoineRondelet: `sn` should be `nf` for consistency. We should stick to `nullifiers`

pyClient/Makefile
```diff
@@ -34,3 +34,7 @@ syntax: ${PROTOBUF_OUTPUT}
 
 test: ${PROTOBUF_OUTPUT}
 	python -m unittest
+
+test_contracts: ${PROTOBUF_OUTPUT}
+	python test/test_contract_base_mixer.py
```
dtebbs: I tried to keep the non-unit tests (those that are commands and require the ethereum RPC host) in `test_commands`.  Could these two be moved into `test_commands` so they cannot every cause any confusion with the unit tests (which `unittest` scans).

zeth-contracts/contracts/BaseMixer.sol
rrtoledo: Fixed.

src/circuits/joinsplit.tcc
```diff
@@ -203,13 +215,19 @@ public:
...
                     FMT(this->annotation_prefix, " input_nullifiers[%zu]", i)));
                 a_sks[i].reset(new libsnark::digest_variable<FieldT>(
-                    pb, 256, FMT(this->annotation_prefix, " a_sks[%zu]", i)));
+                    pb,
+                    ZETH_A_SK_SIZE * 8,
+                    FMT(this->annotation_prefix, " a_sks[%zu]", i)));
                 h_is[i].reset(new libsnark::digest_variable<FieldT>(
-                    pb, 256, FMT(this->annotation_prefix, " h_is[%zu]", i)));
+                    pb,
+                    HashT::get_digest_len(),
```
rrtoledo: 256 change to HashtT::gest_digest_len() as asked offline.

PR #81, opened 2019-09-30
Blake2s review

AntoineRondelet:


AntoineRondelet:
I added comments where needed, used variables/const where necessary and cleaned a few formatting issues.
Please make sure to add comments where necessary to make sure that we can always easily understand the code, especially when using tricks in the circuit

src/circuits/blake2s/blake2s_comp.hpp
```diff
@@ -34,15 +35,16 @@ class BLAKE2s_256_comp : public libsnark::gadget<FieldT>
...
+    std::array<std::vector<g_primitive<FieldT>>, rounds> g_arrays;
     std::vector<xor_constant_gadget<FieldT>> xor_vector;
 
     // TODO: Remove ZERO and pass it in the constructor
     libsnark::pb_variable<FieldT> ZERO;
 
 public:
     std::array<std::array<FieldT, 32>, 8> IV;
-    std::array<std::array<uint, 16>, 10> sigma;
+    std::array<std::array<uint, 16>, 10> sigma; // I think we need to replace 10 by nb_rounds?
```
rrtoledo: Yes we can.

src/circuits/simple_gadgets.tcc
```diff
@@ -70,14 +69,22 @@ xor_constant_gadget<FieldT>::xor_constant_gadget(
...
     for (size_t i = 0; i < a.size(); i++) {
         this->pb.add_r1cs_constraint(
             libsnark::r1cs_constraint<FieldT>(
-                -FieldT("2") * (FieldT("1") - FieldT("2") * c[i]) * a[i],
-                b[i],
-                res[i] - c[i] - a[i] * (FieldT("1") - FieldT("2") * c[i]) -
-                    b[i] * (FieldT("1") - FieldT("2") * c[i])),
+                2 * (a[i] + b[i] - 2 * a[i] * b[i]),
+                c[i],
+                a[i] + b[i] - 2 * a[i] * b[i] + c[i] - res[i]),
```
rrtoledo: The variable c was supposed to be the constants which is the reason why I wrote the formula like this. It seems here that you consider b to be the constant.
I supposed we should write what variable is the constant somewhere in the code not to confuse the reader.

src/circuits/simple_gadgets.tcc
```diff
@@ -68,7 +69,19 @@ xor_constant_gadget<FieldT>::xor_constant_gadget(
...
+    // we can write: res = a + b - 2ab
+    //
+    // Hence, res2 = a XOR b XOR c = (a XOR b) XOR c <=> res XOR c
+    // Thus: res2 = res XOR c is constrainted as:
+    //   2(res) * c = res + c - res2
+    // which leads to:
+    //   2(a + b - 2ab) * c = a + b - 2ab + c - res2
+    // => res2 = 2ac + 2bc - 4abc - a - b + 2ab - c
+    // => res2 = b * [2 * (1 - 2c) * a] + b * (2c - 1) + a * (2c - 1) - c
+    // => res2 - b * (1 - 2c) - a * (1 - 2c) - c = b * [-2(1 - 2c) * a]
```
rrtoledo: There is a double negative error in line 84, we should have (after reordering a bit):
res2 + a*(1-2c) + b*(1-2c) + c = 2ab*(1-2c)

src/circuits/simple_gadgets.tcc
```diff
@@ -25,12 +25,13 @@ xor_gadget<FieldT>::xor_gadget(
...
 {
-    // 32 constraints
+    // Set the constraints (#constraints = length of bit string)
     for (size_t i = 0; i < a.size(); i++) {
+        // res = a XOR b <=> (2.a) * b = a + b - res
         this->pb.add_r1cs_constraint(
             libsnark::r1cs_constraint<FieldT>(
-                -2 * a[i], b[i], res[i] - a[i] - b[i]),
-            FMT(this->annotation_prefix, " rotated_xored_bits_%zu", i));
+                2 * a[i], b[i], a[i] + b[i] - res[i]),
```
rrtoledo: Should we replace 2 by FieldT("2")?

src/circuits/simple_gadgets.tcc
AntoineRondelet: Good catch thanks, hopefully it's fixed now

src/circuits/simple_gadgets.tcc
AntoineRondelet: This should be cast to `FieldT`

PR #86, opened 2019-10-11
WIP: [MPC] mpc coordinator

dtebbs:
POC server which, given some configuration (including a list of participants) allows clients to download the latest challenge and submit responses (contributions) during their time-slot.  As contributions arrive they are applied to the latest accumulator and the new challenge computed.  The next participant is notified that his time-slot has begun, as soon as new challenges are ready or as time slots expire.

Support for phase1 and phase2 servers and clients.


dtebbs:
Temporarily closing

dtebbs:
Reopening after some clean up and refactoring

PR #92, opened 2019-10-22
Updating encoding function (was deprecated)

rrtoledo:
The new double spend test had an error which led to a false negative test result. 

More precisely, the function to encode some primary inputs was deprecated, which led to an exception. This led to potential false positive for the double spending test.


rrtoledo:
Wrong source branch (should be packing)

PR #104, opened 2019-11-01
Cleaned up function to re-assemble primary inputs and added comments

AntoineRondelet:
@rrtoledo here's a starting point to cleanup the packing branch. Please have a look, and add comments where needed to clarify what you do at the end of the function.

### DIFF
--- a/zeth-contracts/contracts/BaseMixer.sol
+++ b/zeth-contracts/contracts/BaseMixer.sol
@@ -172,7 +172,7 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
         // (if residual_bits_length > field_capacity) as such we compute the indices of the
         // first and last field elements they are located at
         //
-        // Here we retreive the indices of the primary inputs containing the bits we want to extract
+        // Here we retrieve the indices of the primary inputs containing the bits we want to extract
         uint first_residual_field_element_index = 1 + 1 + 2 * jsIn + jsOut + (offset + start)/field_capacity;
         uint second_residual_field_element_index = first_residual_field_element_index;
         if (nb_field_residual > 1) {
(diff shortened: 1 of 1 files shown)


rrtoledo:
@AntoineRondelet just added comments about how we reassemble the bytes and discard the unneeded bits.

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -150,36 +163,110 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
+        // Similarly, the binary operation "res >> 16-length" corresponds to dividing by 2**(16 - length)
         res = res / uint16(2**(16 - length));
 
+        // We now have something like this:
+        // 0 || ... || 0 || b_{start} || ... || b_{end}
+        //                  ^                         ^
+        //                  |_________________________|
+        //                               |
+        //                      (length) asked bits
+        // All the asked bits are in the second byte, we thus return this one.
```
AntoineRondelet: > All the asked bits are in the second byte, we thus return this one.

This is only true if `length` is smaller than a byte, and it is just required to be < 16, so it can be more than a byte

rrtoledo: That is an error then. 
The function is supposed to return 1 byte only. If we wanted to return up to two bytes we would have to work on three bytes inside the function.
I propose to require that `length` is smaller than a byte (change the require to <8).

AntoineRondelet: Yes, then we'd need to change the conditional of https://github.com/clearmatics/zeth/blob/packing/zeth-contracts/contracts/BaseMixer.sol#L145-L149

rrtoledo: Updated.
The comment here is not the most clear I think: https://github.com/clearmatics/zeth/blob/aaa80b9d15d68f7c9b60ce60f6c5081beafe6003/zeth-contracts/contracts/BaseMixer.sol#L153-L154

We should say that we will work on at most field elements (not two bytes).
Then we should say that we (arbitrarily) assume that we extract at most one byte (hence the require). As 8 bits (a byte) can be written over at most two bytes, we will work at most on two bytes.

AntoineRondelet: > We should say that we will work on at most field elements (not two bytes)

> As 8 bits (a byte) can be written over at most two bytes, we will work on at most two bytes.

Yes happy to have this changed to say that we're manipulating at most 2 field elements, and that 8 consecutive bits are (at most) written on 2 bytes, if you think it is clearer @rrtoledo

rrtoledo: Updated.
I also added a couple of require on the other inputs.

Unfortunately this led to a `stack too deep`, so I had to remove a couple of variables (but added comments to compensate).

AntoineRondelet:
LGTM, happy for you to merge it to #80 and continue to work there (for the tests and so on)

PR #114, opened 2019-11-08
Code cleanup (pyClient, cpp, proto files)

AntoineRondelet:


dtebbs:
I seem to have accidentally included some libsnark changes.  I'll revert that.

pyClient/setup.py
```diff
@@ -0,0 +1,62 @@
...
+    exit(1)
+
+setup(
+    name='zeth',
+    version='0.1',
+    description='Interface to zeth operations',
+    packages=find_packages(),
+    install_requires=[
+        "mypy==0.720",
+        "mypy-protobuf",
```
dtebbs: mising version

pyClient/zeth/joinsplit.py
```diff
@@ -0,0 +1,723 @@
...
+
+# Dictionary representing a VerificationKey from any supported snark
+GenericVerificationKey = Dict[str, Any]
+
+
+# Dictionary representing a Proof from any supported snark
+GenericProof = Dict[str, Any]
+
+
+# class ZkSnarkProvider(object):
```
dtebbs: I'll remove this.

.gitignore
```diff
@@ -1,39 +1,37 @@
...
-pyClient/util_pb2.py
-pyClient/util_pb2_grpc.py
-
-# Ignore pycash
-pyClient/__pycache__/
-
+/zeth-contracts/node_modules/*
+
+# Ignore pyClient env and build directories
+/pyClient/zeth-devenv/
```
AntoineRondelet: The venv has been renamed to `env`

pyClient/test_commands/test_mimc_hash.py
```diff
@@ -0,0 +1,71 @@
+import zeth.contracts as contracts
+
+
+def test_mimc() -> None:
+    """
+    Benchmarking Gas cost for different MiMC exponents
+    """
+
+    # compile MiMC, MerkleeTree contracts
```
AntoineRondelet: `MerkleeTree` -> MerkleTree

pyClient/zeth/joinsplit.py
```diff
@@ -0,0 +1,723 @@
...
+G1 = Tuple[ec.FQ, ec.FQ]
+
+
+class ApkAskPair:
+    def __init__(self, a_sk: str, a_pk: str):
+        self.a_pk = a_pk
+        self.a_sk = a_sk
+
+
+SK = Tuple[int, int]
```
AntoineRondelet: Why not, `SK = Tuple[int, int]` -> `SK = Tuple[FQ, FQ]`?

pyClient/zeth/joinsplit.py
```diff
@@ -0,0 +1,723 @@
...
+VK = Tuple[G1, G1]
```
AntoineRondelet: I'd rename `VK` to something else as we also use `VK` in the proto files, and in other places so it adds confusion. What about `JS_PK` (for JoinsplitPubKey), and `SK` -> `JS_SK` (for JoinsplitSecretKey)?

pyClient/zeth/joinsplit.py
```diff
@@ -0,0 +1,723 @@
...
+        self.rho = rho
+        self.trap_r = trap_r
+
+    @staticmethod
+    def new() -> NoteRandomness:
+        """
+        Compute the note randomness: the trapdoor trapR and rho. Starting the
+        Non-Malleability update, rho is computed from phi (see above), the rho
+        generated in this function is thus obsolete except for dummy input
+        notes.
```
AntoineRondelet: I'm not sure it is a good idea to stick to this function since rho needs to be derived in another way (from phi as mentioned in the comment)

pyClient/setup.py
dtebbs: Added

pyClient/zeth/joinsplit.py
```diff
@@ -0,0 +1,680 @@
...
+        value0: str,
+        recipient_apk1: str,
+        value1: str) -> Tuple[ZethNote, ZethNote]:
+    """
+    Create two ordered ZethNotes. This function is used to generate new output
+    notes.
+    """
+    rho0 = compute_rho_i(phi, hsig, 0)
+    randomness0 = NoteRandomness.new()
+    note0 = ZethNote(
```
AntoineRondelet: I think some refactoring will be needed here. We can modify the `create_zeth_note` function and invoke it in this one

pyClient/zeth/joinsplit.py
```diff
@@ -0,0 +1,680 @@
...
+        rho=rho1,
+        trap_r=randomness1.trap_r
+    )
+
+    return note0, note1
+
+
+def parse_zeth_note(zeth_note_grpc_obj: ZethNote) -> Dict[str, str]:
+    note_json = {
+        "aPK": zeth_note_grpc_obj.apk,
```
AntoineRondelet: Let's switch to snake_case for the key names

dtebbs:
> I seem to have accidentally included some libsnark changes. I'll revert that.
Reverted this

pyClient/zeth/mimc.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+def keccak_256(data: bytes) -> keccak.Keccak_Hash:
+    return keccak.new(data, digest_bits=256)
+
+
+class MiMC7:
+    def __init__(
+            self,
+            seed: str = "clearmatics_mt_seed",
```
AntoineRondelet: Seems like this seed should be added to the constants file (named something like `MIMC_MT_SEED`)

pyClient/zeth/joinsplit.py
dtebbs: Yeah, I was wondering what to do about this.
I've removed the `NoteRandomness` and now just have a `trap_r_randomness()` that returns a random `r`.
All the obsolete stuff is moved to `mock.py` since that's the only place it is used.

.gitignore
dtebbs: Thanks.  Fixed.

pyClient/test_commands/test_mimc_hash.py
dtebbs: Should be fixed in the latest.

pyClient/zeth/joinsplit.py
dtebbs: Changed

pyClient/zeth/joinsplit.py
dtebbs: I renamed these to `JoinsplitPUblicKey` etc (to match the other types like `JoinsplitInput`).  Easy to rename them to somethign else if you prefer.

AntoineRondelet: Perfect, thanks

pyClient/zeth/mimc.py
dtebbs: Updated this and moved all the test code out of here into the `test` module.

pyClient/zeth/joinsplit.py
dtebbs: Ididn't see this until I'd already refactored as above.  Let me know if that works.

pyClient/zeth/joinsplit.py
dtebbs: Updated

pyClient/zeth/joinsplit.py
AntoineRondelet: We can keep it that way (ie: remove the `create_zeth_note` as you did). What I was suggesting was to create each notes in this function by calling the `create_zeth_note`, but what you did works, let's keep it that way.

PR #123, opened 2019-11-18
Fix pghr13

dtebbs:
Fixes a few problems with the PGHR13 build, now that some MPC functionality is only available for GROTH16.  (Thanks @rrtoledo for catching this).
Also added the PGHR13 config to the travis build.


.travis.yml
```diff
@@ -36,5 +36,8 @@ matrix:
         - env: CI_TASK=build CI_CONFIG=Debug
           os: osx
           osx_image: xcode11
+        - env: CI_TASK=build CI_CONFIG=Release CI_ZKSNARK=PGHR16 CI_USE_DOCKER=1
```
AntoineRondelet: `PGHR16` -> PGHR13

dtebbs: Thanks!  Actually, that should have caused an error in cmake on travis.  Turns out I need to explitily pass the envvar to docker.  Added that as well.

AntoineRondelet: Yes was surprised to see the build passing

PR #124, opened 2019-11-18
Fixed helper debug functions

AntoineRondelet:


debug/analyzer/parse_r1cs.py
```diff
@@ -0,0 +1,158 @@
...
+
+        ...
+
+        {
+            "index": INDEX,
+            "annotation": ANNOTATION
+        }
+    ]
+"""
+def get_index(annotation_set, annotation):
```
dtebbs: Contrary to many languages, Python documentation comments usually go immediately inside the function, right at the start.  (The runtime then associates the text with `get_index.__doc__`, I think.)

src/prover_server/prover_server.cpp
```diff
@@ -312,12 +312,16 @@ int main(int argc, char **argv)
...
-        std::cout << "[DEBUG] Generate new keypair" << std::endl;
+        std::cout << "[INFO] Generate new keypair" << std::endl;
         return prover.generate_trusted_setup();
     }();
 
-    std::cout << "[DEBUG] Setup successful, starting the server..."
-              << std::endl;
+#ifdef DEBUG
+    std::cout << "[DEBUG] Dump R1CS to json file" << std::endl;
+    prover.dump_constraint_system();
```
dtebbs: Out of curiosity, does this take a long time? If it does, we might consider doing it under control of a command line flag rather than `DEBUG`.  Not necessary, just a thought for the future.

debug/analyzer/parse_r1cs.py
AntoineRondelet: Good one, of course, thanks @dtebbs will move the comments down

src/prover_server/prover_server.cpp
AntoineRondelet: It does slow the server at starting time a bit. Happy to see this move to a flag `--debug` for the prover_server. Note however, that the annotations I use are only avail in debug mode: https://github.com/clearmatics/libsnark/blob/master/libsnark/relations/constraint_satisfaction_problems/r1cs/r1cs.hpp#L126-L129 so it make sense to keep it that way, at least now. We can think about introducing more changes across libraries later on, but for now having a decent export of the r1cs to json is helpful I think

dtebbs: OK.  I didn't realise the annotations are only available in DEBUG.  I guess that makes sense, though.  If it's 10-15s, it could make sense to have a `--dump-r1cs` flag in the future, but I guess not necessary for now.

Issue #8, opened 2019-03-28
Create a CLI to facilitate user experience


AntoineRondelet:

dannywillems:
Can you give a bit more info pls? What do you want exactly?

AntoineRondelet:
Sure, thanks for the message @dannywillems!

As of now, the "Proof Consumer" (PC for short) is just implemented as a bunch of Python scripts (see the `pyClient` folder). These scripts execute a "mock" scenario with Alice, Bob and Charlie doing Ether or ERC20 transfers in zero knowledge (the scripts call the "Proof Generator" via RPC calls and use the received proofs to call the "Mix" function of the Mixer smart contract, to effectively do private transfers).
In this issue, we propose to enhance usability of zeth by having a better way to interact with the Proof Generator and the Mixer smart contract to do private asset transfer. In fact, rather than writing your script describing the scenario you'd like to test; it'd be great to have a CLI (or even better a UI) that enables the user to enter the payment informations, then call the PG on the given inputs to get a proof and then call the Mixer to execute the transfers.

AntoineRondelet:
As mentioned on the comment above, the purpose of this issue is to have a basic CLI to enable users to do payments via the Zeth contract.
Thus, the basic functionalities desired are:
- `Payment`, something like `zeth pay [options and args]`
- `Receive` a payment `zeth receive [options and args]`, see https://github.com/clearmatics/zeth/issues/11 (note we logged the addresses of the inserted commitments to ease this check). Here we basically need to listen to the mixer events to try to decrypt the ciphertexts logged (are we the recipients?) + we need to recompute the commitment(s) and check it against the one(s) appended in the merkle tree to make sure the payment is valid. Note, this command should be quite similar to the one to run a note recovery procedure (imagine a hardware failure that yields a loss of the content of the coin store), something like `zeth scan --keyfile [yourViewingKeyFile] --depth [depthOfTheScan]`. This would scan the blockchain up to the specified depth and try to repopulate the coinstore (assume that the content of the `keystore` was backed up). If a payment is successfully received, then we write the note in a file in the coinstore (like already done in the test scripts)

AntoineRondelet:
Closing this issue since the corresponding PR has been merged

PR #130, opened 2019-12-02
Zeth command line tools

dtebbs:
- Some further reefactoring into high level functions
- Command line tools or basic deposit, spend withdraw operations
- Some other tools for convenience / demos


dtebbs:
Updated the README files.  Worth looking at these (final commit) to get a good overview of what the tools do and how they work.

pyClient/README.md
```diff
@@ -45,11 +55,192 @@ and fails to fetch it (or fails to fetch the right version).
...
 
-To run this command you need to replace the solidity version (denoted by
-`[solc-version]`), the python version (denoted `[python-version]`), and binary
-file (denoted by `[solc-for-your-distribution]`) by your system specific pieces
-of information.
+To run this command, replace the solidity version (denoted by `[solc-version]`),
+the python version (denoted by `[python-version]`), and binary file (denoted by
+`[solc-for-your-distribution]`) by your system specific information.
+
+# The `zeth` command
```
AntoineRondelet: "The `zeth` command line interface"

pyClient/README.md
```diff
@@ -45,11 +55,192 @@ and fails to fetch it (or fails to fetch the right version).
...
+- interact with an Ethereum RPC host
+- interact with the deployed Zeth contracts
+- request proofs and keys from the `prover_server`, and
+- access secret and public data for the current user
+
+Some of this data could not easily be specified by command line, so the `zeth`
+tools assume it can be found in specific files (with default file names,
+overridable on the command line).
+
+The set of files required for by zeth for a single user is described below.  We
```
AntoineRondelet: `required for by zeth for a` -> required by zeth for a

pyClient/README.md
```diff
@@ -45,11 +55,192 @@ and fails to fetch it (or fails to fetch the right version).
...
+$ mkdir alice
+$ cd alice
+
+# Specify an eth-address file for an (unlocked) Ethereum account
+$ echo 0x.... > eth-address
+
+# Copy the instance file (received from the deployer)
+$ cp <shared-instance-file> zeth-instance.json
+
+# Generate new secret (zeth-key.json)and public key (zeth-key.json.pub)
```
AntoineRondelet: > Generate new secret (zeth-key.json)and

Generate new secret (zeth-key.json) and

pyClient/README.md
```diff
@@ -45,11 +55,192 @@ and fails to fetch it (or fails to fetch the right version).
...
+
+With these files in place, `zeth` commands invoked from inside this directory
+can perform actions on behalf of Alice.  We call this Alice's *client directory*
+below, and assume that all commands are executed in a directory with these
+files.
+
+## Depositing funds
+
+Alice can deposit funds from the address in `eth-address` into the Zeth
+instance, as a private Zeth note, spendable with her secret key `zeth-key.json`:
```
AntoineRondelet: Zeth instance -> Zeth mixer instance

pyClient/README.md
```diff
@@ -45,11 +55,192 @@ and fails to fetch it (or fails to fetch the right version).
...
+
+```console
+# Deposit 10 ether, creating a Zeth note owned by Alice
+(env)$ zeth deposit 10
```
AntoineRondelet: May be worth adding a flag to be able to use ERC tokens @dtebbs 

Something like:
```console
zeth deposit 10 --token <erc-token-contract-address>
```

If no flag is given, it defaults to ETH

pyClient/README.md
```diff
@@ -45,11 +55,192 @@ and fails to fetch it (or fails to fetch the right version).
...
+```
+
+Note that for this deposit to be reflected in Alice's client direectory, she
```
AntoineRondelet: direectory -> directory

pyClient/README.md
```diff
@@ -45,11 +55,192 @@ and fails to fetch it (or fails to fetch the right version).
...
+```console
+$ zeth notes
+b1a2feaf: value=200, addr=0
+eafe5f84: value=100, addr=2
+
+$ zeth spend \
+    --in eafe5f84 \                       # "eafe5f84: value=100, addr=2"
+    --in 0 \                              # "b1a2feaf: value=200, addr=0"
+    --out d77f...0e00:cc7c....7f76,120 \  # 120 to this addr
+    --out 3a43...fd3b:9fc8....b838,180    # 180 to this addr
```
AntoineRondelet: Do we have a way to specify `v_in` to deposit funds?

I'd be in favor of merging all of these commands (`spend`, `withdraw` and so on) into a single command, like `zeth pay` (as in: https://github.com/clearmatics/zeth/issues/8#issuecomment-555618678), or `zeth mix` (to remain consistent with the function name on the mixer).

Separating `deposits`, `withdrawals` and so on, is the type of anti-patterns that we should avoid at any cost (at all stages of the project) as it leaks a lot of *exact* data to an adversary. Sticking to a single functionality would enable users to deposit, withdraw, transfer notes into a single command which provides more flexibility and which is better as it adds noise (the adversary does not exactly know what the user is doing, ie: the user may be able to withdraw and transfer in the same transaction, and so on).

To do so, we'd need, something like:
`zeth mix --in <note1> --in <note2> --vin <v_in> --vout <v_out> --out <a_pk1> --out <a_pk2>`
(we can use default values if the flags are not set)

pyClient/README.md
```diff
@@ -45,11 +55,192 @@ and fails to fetch it (or fails to fetch the right version).
...
+Ethereum network.  By default, these operations output the transaction ID that
+was broadcast.  This Id can be passed to `zeth sync` via the `--wait-tx` flag,
+forcing this command to wait for that transaction to be committed before
+searching for new notes.  Alternatively, the `--wait` flag can be passed to the
+`deposit`, `spend` and `withdraw` operations to make them wait and sync new
+notes before exiting.
+
+## Limitations - Note and Key management
+
+As proof-of-concept softare, these tools are not suitable for use in a
```
AntoineRondelet: software

pyClient/README.md
```diff
@@ -45,11 +55,192 @@ and fails to fetch it (or fails to fetch the right version).
...
+the client tools and is considered out of scope for this proof-of-concept.  The
+user must manually track which notes have been spent (for example by moving
+their files into a `spent` subdirectory where they will not be seen by the
+wallet).
+
+All values that make up the zeth secret key, and secrets (required to spend
+notes) are stored in plaintext.  A fully secure client would encrypt these to
+protect them from malicious entities that may gain access to the file system.
+Such client-side security mechanisms are also beyond the scope of this
+proof-of-concept implementation.
```
AntoineRondelet: You can also add that backups for the keys would be necessary but outside of the scope as this project is still in PoC stage.

pyClient/README.md
dtebbs: Changed

pyClient/README.md
dtebbs: Removed

pyClient/README.md
dtebbs: Fixed.  Thanks.

pyClient/README.md
dtebbs: Updated

pyClient/README.md
dtebbs: Fixed.  Thanks.

pyClient/README.md
dtebbs: Fixed.

pyClient/README.md
dtebbs: Added a note about this.

pyClient/README.md
dtebbs: I've added a `--token-address` flag to the `deploy` command.  The token address gets saved alongn with the zeth mixer instance data and used in the `mix` operations.  I've also added a `zeth token_approve` command for dealing with token.
Finally, a `deploy_test_token` command (in `test_commands`, not part of the main cli interface) allows the minimal token implementation to be deployed for demos / testing.

pyClient/README.md
dtebbs: Replaced `deposit`, `withdraw` and `send` with a single `mix` function.  User must compute input and output values correectly, or the command throws an error.

README.md
```diff
@@ -36,35 +41,44 @@ docker build -f Dockerfile-zeth -t zeth-dev .
...
 make
+# (optional) Run the unit tests
+make test
+# (optional) Run the all tests (unit tests, syntax checks, etc)
+make check
 
-# Start the proving server
-prover_server
+# Start the prover_server process
+src/prover_server
```
AntoineRondelet: why `src/prover_server`? The path to the binaries is set by `. ./setup_env.sh`

README.md
```diff
@@ -36,35 +41,44 @@ docker build -f Dockerfile-zeth -t zeth-dev .
...
 ```
 
-Note: *flags* in the cmake command may include `-DCMAKE_BUILD_TYPE=Release` for
-an optimized build.
+Note: By default, the prover_server generates a key at startup.  Flags can be
```
AntoineRondelet: Please use single spaces after punctuation `.`
i.e. `startup.<space><space>Flags` -> `startup.<space>Flags`

AntoineRondelet: prover_server -> `prover_server` (let's use `` since prover_server is a binary/command here)

README.md
```diff
@@ -80,22 +94,30 @@ npm install
 npm run testrpc
 ```
 
-#### Terminal 3: Start the testing Python stub
+#### Terminal 3: Python tools
```
AntoineRondelet: Python client

README.md
```diff
@@ -80,22 +94,30 @@ npm install
...
 
 cd pyClient
-# Follow the few steps described in the README of the python stub
 ```
 
-### Use the pyClient
+Follow the steps described in the [pyClient README](pyClient/README.md) to run
+tests or invoke the zeth tools.
+
+### pyClient and jsClient
```
AntoineRondelet: This seems to be a old version of the README, the `jsClient` has been removed and all mentions of it removed from the README (see: https://github.com/clearmatics/zeth/#use-the-pyclient)

debug/analyzer/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright (c) 2015-2019 Clearmatics Technologies Ltd
```
AntoineRondelet: Extraneous line.

I didn't include the copyright headers to the `__init__.py` files since there are basically empty files. I'd add `#!/usr/bin/env python3` though

pyClient/README.md
```diff
@@ -45,11 +55,211 @@ and fails to fetch it (or fails to fetch the right version).
...
+brief description is given in this section.  More details are available via
+`zeth --help`, and example usage can be seen in the [pyclient test
+script](../scripts/test_zeth_cli).
+
+## Environment
+
+Depending on the operation being performed, the `zeth` client must:
+- interact with an Ethereum RPC host
+- interact with the deployed Zeth contracts
+- request proofs and keys from the `prover_server`, and
```
AntoineRondelet: Please specify that it is the Verification key (otherwise it can get confusing)

pyClient/README.md
```diff
@@ -45,11 +55,211 @@ and fails to fetch it (or fails to fetch the right version).
...
+Specify the note(s) to be withdrawn, and the total value as public output:
+```console
+$ zeth withdraw --in eafe5f84 --vout 100
+```
+
+### Complex Transactions
+
+Public and private inputs and outputs can all be specified for a single
+transaction.  Combining multiple operations can help to obfuscate the meaning of
+the transaction.
```
AntoineRondelet: I'd rephrase this section and change the name to something like "A note on the `zeth mix` command", then explain why there is a single command (which seems to be what you do here). 

Maybe something along the lines of:
```
### A note on the `zeth mix` command

As explained above, the `zeth mix` command can be used to deposit funds on the mixer, transfer notes, and withdraw funds from the mixer. Using a single command to achieve all these purpose allows to combine all these operations at will, which greatly improves the privacy level provided by Zeth. In fact, no exact information about the meaning of a transaction is ever leaked to the an observant attacker.

Here are a few examples of complex payments allowed by the Mix function:

$ zeth notes
b1a2feaf: value=200, addr=0
eafe5f84: value=100, addr=2

$ zeth mix \
    --in eafe5f84 \                  # "eafe5f84: value=100, addr=2"
    --vin 5
    --out d77f...0e00:cc7c....7f76,120 \  # 103 to this addr (Bob's address)
    --out 3a43...fd3b:9fc8....b838,180    # 2 to this addr (Someone else's address - can be mine if I need refund)
```

pyClient/commands/zeth
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env python3
+
```
AntoineRondelet: Add copyright header please

pyClient/test_commands/test_erc_token_mixing.py
```diff
@@ -264,36 +188,31 @@ def main() -> None:
         bob_eth_address,
         alice_eth_address,
         charlie_eth_address,
-        mixer_instance.address
+        zeth_client.mixer_instance.address
     )
 
     # Charlie tries to decrypt the notes from Bob's previous transaction.
-    notes_charlie = charlie_wallet.receive_notes(
+    note_descs_charlie = charlie_wallet.receive_notes(
```
AntoineRondelet: what does `descs` stand for? description?

pyClient/zeth/contracts.py
```diff
@@ -4,40 +4,77 @@
...
+        return InstanceDescription(instance.address, instance.abi)
+
+    def instantiate(self, web3: Any) -> Any:
+        """
+        Return the instantiated contract
+        """
+        return web3.eth.contract(address=self.address, abi=self.abi)
+
+
+def get_block_number(w3: Any) -> int:
```
AntoineRondelet: let's use either `w3` everywhere or stick to `web3` to be consistent throughout the code

pyClient/zeth/joinsplit.py
```diff
@@ -45,6 +52,32 @@
 ComputeHSigCB = Callable[[bytes, bytes, JoinsplitSigVerificationKey], bytes]
 
 
+def blake2s_compress(left: bytes, right: bytes) -> bytes:
+    """
+    Execute blake2s as a compression function, ensuring that the input is of the
+    correct length.  (Note we support len(left) != len(right) as long as the
+    total length is appropriate).
```
AntoineRondelet: total length is appropriate -> total input length is appropriate

By appropriate you mean, input length = block size, right?

pyClient/zeth/joinsplit.py
```diff
@@ -53,6 +86,26 @@ def __init__(self, a_pk: OwnershipPublicKey, k_pk: EncryptionPublicKey):
         self.a_pk: OwnershipPublicKey = a_pk
         self.k_pk: EncryptionPublicKey = k_pk
 
+    def __str__(self) -> str:
+        """
+        Write the address as "<ownership-key-hex>:<encryption_key_hex>".
+        (Technically the ":" separator is not required, since the first key is
+        written with fixed length, but it is provides a small amount amount of
+        sanity checking).
```
AntoineRondelet: but it is provides a small amount amount of sanity checking -> can you rephrase this?

pyClient/zeth/joinsplit.py
```diff
@@ -316,30 +394,114 @@ class ZethClient:
...
+        (mixer_instance, initial_merkle_root) = contracts.deploy_contracts(
+            web3,
+            mk_tree_depth,
+            proof_verifier_interface,
+            otsig_verifier_interface,
+            mixer_interface,
+            hasher_interface,
+            vk_json,
+            deployer_eth_address,
+            4000000,
```
AntoineRondelet: Let's pass the deployment gas as an argument to this function (we can use a default value to `4000000` to keep the tests unaffected)

pyClient/zeth/wallet.py
```diff
@@ -6,40 +6,193 @@
...
+        return ZethNoteDescription(
+            note=joinsplit.zeth_note_from_json_dict(json_dict["note"]),
+            address=int(json_dict["address"]),
+            commitment=bytes.fromhex(json_dict["commitment"]))
+
+
 class Wallet:
-    def __init__(self, username: str, wallet_dir: str, k_sk_receiver: PrivateKey):
+    """
+    Very simple class to track a list of notes addressed to some owner.  Note
```
AntoineRondelet: `Very simple class to track a list of notes addressed to some owner` -> `Very simple class to track the list of notes owned by a user of Zeth`

pyClient/zeth/wallet.py
```diff
@@ -6,40 +6,193 @@
...
+    this does not store the notes in encrypted form, and encodes some
+    information (invluding value) in the filename.  It is NOT intended to be
```
AntoineRondelet: invluding -> including

debug/analyzer/__init__.py
dtebbs: I've generally been putting the she-bang into files that might be run as shell commands, esecially if they don't have the `.py` extension, since it's also a hint to editors.

Are you suggesting we enforce it in all python files?  If so, I'll add a little script to enforce consistency.

pyClient/README.md
```diff
@@ -45,11 +55,211 @@ and fails to fetch it (or fails to fetch the right version).
...
+Some examples are given below
+
+### Depositing funds
+
+A simple deposit consists of some public input (ether or tokens), and the
+creation of a private note.
+
+```console
+# Deposit 10 ether from `eth-address`, creating a Zeth note owned by Alice
+(env)$ zeth mix --out <public_key>,10 --vin 10
```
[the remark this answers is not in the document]
dtebbs: Yes.  For the `mix` command, the values of the input notes must be given explicitly.

pyClient/test_commands/test_erc_token_mixing.py
dtebbs: Yes.  Previously this was just the note data, but now it is a fuller description (including the "address" in the merkle tree) currently required to spend the note.

README.md
dtebbs: reverted

README.md
dtebbs: changed

README.md
dtebbs: Must have been a rebasing issue.  I've removed the section.

pyClient/README.md
```diff
@@ -25,14 +32,17 @@ $ source env/bin/activate
 
 ## Execute testing client
 
+These are scripts that perform some predetermined set of transactions between
+dummy clients Alice, Bob and Charlie.
```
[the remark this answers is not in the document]
dtebbs: changed

pyClient/README.md
dtebbs: Changed to 'proof verification key'

pyClient/README.md
```diff
@@ -45,11 +55,211 @@ and fails to fetch it (or fails to fetch the right version).
...
+**Public Input.** Ether or ERC20 token value send to the mixer.
+
+**Public Output.** Ether or ERC20 tokens value to be extracted from the mixer.
+
+Some examples are given below
+
+### Depositing funds
+
+A simple deposit consists of some public input (ether or tokens), and the
+creation of a private note.
```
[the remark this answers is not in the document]
dtebbs: changed

pyClient/README.md
```diff
@@ -45,11 +55,211 @@ and fails to fetch it (or fails to fetch the right version).
...
+## Environment
+
+Depending on the operation being performed, the `zeth` client must:
+- interact with an Ethereum RPC host
+- interact with the deployed Zeth contracts
+- request proofs and keys from the `prover_server`, and
+- access secret and public data for the current user
+
+Some of this data could not easily be specified by command line, so the `zeth`
+tools assume it can be found in specific files (with default file names,
```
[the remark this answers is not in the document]
dtebbs: reworded

pyClient/README.md
```diff
@@ -45,11 +55,211 @@ and fails to fetch it (or fails to fetch the right version).
...
+overridable on the command line).
+
+The set of files required by zeth for a single user is described below.  We
+recommend creating a directory for each user, containing these files.  In this
+way, it is very easy to setup one or more conceptual "users" and invoke `zeth`
+operations on behalf of each of them to experiment with the system.
+
+- `eth-address` specifies an eth address from which to fund transactions.  When
```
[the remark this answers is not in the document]
dtebbs: rewordd

pyClient/README.md
```diff
@@ -45,11 +55,211 @@ and fails to fetch it (or fails to fetch the right version).
...
+  running the testnet (as described in the [top-level README](../README.md)),
+  addresses are created at startup and written to the console.  One of these can
+  be copy-pasted into this file.
+- `zeth-instance.json` contains the address and interface for an instance of the
+  zeth contracts.  This file is created by the deployment step below and should
+  be distributed to each client.
+- `zeth-key.json` and `zeth-key.json.pub` hold the secret and public Zeth keys
+  for the client.  These can be generated with the `zeth key` command.
```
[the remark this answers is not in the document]
dtebbs: Renamed to `zeth gen-address`

pyClient/README.md
```diff
@@ -45,11 +55,211 @@ and fails to fetch it (or fails to fetch the right version).
...
+way, it is very easy to setup one or more conceptual "users" and invoke `zeth`
+operations on behalf of each of them to experiment with the system.
+
+- `eth-address` specifies an eth address from which to fund transactions.  When
+  running the testnet (as described in the [top-level README](../README.md)),
+  addresses are created at startup and written to the console.  One of these can
+  be copy-pasted into this file.
+- `zeth-instance.json` contains the address and interface for an instance of the
+  zeth contracts.  This file is created by the deployment step below and should
+  be distributed to each client.
```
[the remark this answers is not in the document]
dtebbs: No, `zeth-instance.json` contains the info about the instance that the given client shoudl itneract with (the instance whose mekle tree contains the notes in the `notes` dir).  We can't use a single "client" (collection of config files, addresses and notes) across multiple instances.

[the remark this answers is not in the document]
dtebbs: Rewored to try and make this clearer.

pyClient/zeth/joinsplit.py
dtebbs: reworded

pyClient/zeth/joinsplit.py
dtebbs: rephrased this

pyClient/zeth/wallet.py
dtebbs: rephrased

pyClient/zeth/wallet.py
dtebbs: fixed

pyClient/zeth/joinsplit.py
dtebbs: Added a `deploy_gas` argument.

debug/analyzer/__init__.py
AntoineRondelet: > I've generally been putting the she-bang into files that might be run as shell commands

Right, let's keep it that way. Adding the hashbang to all (not necessarily executable) files in contrary to the intended use.

pyClient/README.md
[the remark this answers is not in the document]
AntoineRondelet: > No, zeth-instance.json contains the info about the instance that the given client shoudl itneract with (the instance whose mekle tree contains the notes in the notes dir).

Aren't we saying the same thing here? By "Do you mean that it contains the address of all deployed Zeth contracts and their ABIs?" I meant that the ABIs of the deployed Mixer, Merkle Tree, Verifier contracts (and their respective addresses) **with which the client will interact** are all written in the `zeth-instance.json` file. In other words, the client only talks to a given deployed `Mixer` contract instance, but there are other contracts that need to be deployed on-chain, such as the MKTree, Verifier, ERC token (if applicable) contracts.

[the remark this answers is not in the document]
dtebbs: > Hum... aren't we saying the same thing here? ...

Ah, OK.  I thought you meant something else.  
`zeth-instance.json` contains the mixer details (address and ABI) and the same for the token, if one is present.  It does not contain the data about the other contracts (verifiers, hashers etc) because AFAICT, the client does not interact with these directly (and also in the optimization PR I've turned some of these into libraries to help save gas during `mix` calls).

pyClient/README.md
```diff
@@ -45,11 +55,211 @@ and fails to fetch it (or fails to fetch the right version).
...
+# in the ./notes directory.
+(env)$ zeth sync
+```
+
+Any notes found are stored in the `./notes` directory, as files containing the
+secret data required to spend them.  This forms a very primitive Zeth "wallet".
+
+```console
+# List all notes received by this client
+$ zeth notes
```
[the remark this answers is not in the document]
dtebbs: renamed the command

pyClient/README.md
```diff
@@ -45,11 +55,211 @@ and fails to fetch it (or fails to fetch the right version).
...
+```
+lists information about all known notes belonging to the current user.
+
+## Mix command
+
+The `zeth mix` command is used to interact with a deployed Zeth instance.  The
```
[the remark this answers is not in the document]
dtebbs: changed

pyClient/zeth/contracts.py
dtebbs: changed everywhere

pyClient/commands/zeth
dtebbs: Added copyright header (and updated script to catch these)

pyClient/README.md
dtebbs: Rephrased

pyClient/README.md
```diff
@@ -45,11 +55,211 @@ and fails to fetch it (or fails to fetch the right version).
...
+eafe5f84: value=100, addr=2
+
+$ zeth mix \
+    --in eafe5f84 \                       # "eafe5f84: value=100, addr=2"
+    --in 0 \                              # "b1a2feaf: value=200, addr=0"
+    --out d77f...0e00:cc7c....7f76,120 \  # 120 to this addr
+    --out 3a43...fd3b:9fc8....b838,180    # 180 to this addr
+```
+
+### Withdrawing value from notesa
```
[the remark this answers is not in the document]
dtebbs: changed

pyClient/README.md
```diff
@@ -45,11 +55,211 @@ and fails to fetch it (or fails to fetch the right version).
...
+
+Specify the note(s) to be withdrawn, and the total value as public output:
+```console
+$ zeth withdraw --in eafe5f84 --vout 100
```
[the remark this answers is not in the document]
dtebbs: fixed

pyClient/README.md
```diff
@@ -45,11 +55,211 @@ and fails to fetch it (or fails to fetch the right version).
...
+# Deposit 10 ether from `eth-address`, creating a Zeth note owned by Alice
+(env)$ zeth mix --out <public_key>,10 --vin 10
+```
+where `<public-key>` is the contents of `zeth-key.json.pub`.
+
+### Privately send a ZethNote to another user
+
+To privately transfer value within the mixer, no public input / output is
+required.  Unspent notes (inputs) and destination addresses and values are
+specified.
```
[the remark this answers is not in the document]
dtebbs: changed

pyClient/README.md
```diff
@@ -45,11 +55,211 @@ and fails to fetch it (or fails to fetch the right version).
...
+
+Some examples are given below
+
+### Depositing funds
+
+A simple deposit consists of some public input (ether or tokens), and the
+creation of a private note.
+
+```console
+# Deposit 10 ether from `eth-address`, creating a Zeth note owned by Alice
```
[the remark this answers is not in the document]
dtebbs: changed

pyClient/README.md
```diff
@@ -45,11 +55,211 @@ and fails to fetch it (or fails to fetch the right version).
...
+via `zeth notes`.  Either the integer "address" or the truncated commitment
+value (8 hex chars) can be used to specify which notes to use as inputs.
+
+**Output Notes.** Given as pairs of Zeth public key and value, separated by a
+comma `,`.  The form of the public key is exactly as in the `zeth-key.json.pub`
+file.  That is, two 32 byte hex values separated by a colon `:`.
+
+**Public Input.** Ether or ERC20 token value send to the mixer.
+
+**Public Output.** Ether or ERC20 tokens value to be extracted from the mixer.
```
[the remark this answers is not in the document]
dtebbs: changed

pyClient/README.md
```diff
@@ -45,11 +55,211 @@ and fails to fetch it (or fails to fetch the right version).
...
+
+**Input Notes.** Zeth notes owned by the current client, which should be visible
+via `zeth notes`.  Either the integer "address" or the truncated commitment
+value (8 hex chars) can be used to specify which notes to use as inputs.
+
+**Output Notes.** Given as pairs of Zeth public key and value, separated by a
+comma `,`.  The form of the public key is exactly as in the `zeth-key.json.pub`
+file.  That is, two 32 byte hex values separated by a colon `:`.
+
+**Public Input.** Ether or ERC20 token value send to the mixer.
```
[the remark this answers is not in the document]
dtebbs: changed

pyClient/README.md
```diff
@@ -45,11 +55,211 @@ and fails to fetch it (or fails to fetch the right version).
...
+- `zeth-key.json` and `zeth-key.json.pub` hold the secret and public Zeth keys
+  for the client.  These can be generated with the `zeth key` command.
+  `zeth-key.json.pub` represents the public address of the client.  This can be
+  shared freely, allowing other users to privately transfer funds to this
+  client.  The secret `zeth-key.json` should **not** be shared.
+
+## Deployment
+
+Deployment compiles and deploys the contracts and initializes them with
+appropriate data to create a new empty instance of the Zeth mixer.  It requires
```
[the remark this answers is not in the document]
dtebbs: reworded

README.md
dtebbs: -> `Python client`

pyClient/README.md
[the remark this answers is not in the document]
AntoineRondelet: Why not call the Mixer contract by its name and:
1. Rename `zeth-instance.json` to `mixer-instance.json` (containing the ABI and address of the Mixer contract deployed)
2. Rephrase to "`mixer-instance.json` contains the address and ABI of the deployed Mixer contract"

If needed, ie: if one wants to use a token, we can still have another file `token-instance.json` that is used if the `--token` flag is set to the mix function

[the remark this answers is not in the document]
AntoineRondelet: these `*-instance.json` files can be stored in a `instances` folder under each client's repository

I think in the end we could have something like:
```
Alice client repo
|_ Ether
|    |_ notes
|    |_ keys
|    |_ config
|        |_ mixer-instance.json
|_ ERCToken1
|    |_ notes
|    |_ keys
|    |_ config
|        |_ mixer-instance.json
|        |_ token-instance.json
|_ ERCToken2
|    |_ notes
|    |_ keys
|    |_ config
|        |_ mixer-instance.json
|        |_ token-instance.json
|_ ...
```

Where we have a mixer per asset being mixed, and where the list of asset folders represent the list of supported assets (Ether and a list of ERC20/223 tokens) by the wallet `

pyClient/zeth/joinsplit.py
```diff
@@ -316,30 +394,117 @@ class ZethClient:
...
+
+    @staticmethod
+    def deploy(
+            web3: Any,
+            prover_client: ProverClient,
+            mk_tree_depth: int,
+            deployer_eth_address: str,
+            zksnark: IZKSnarkProvider,
+            token_address: Optional[str] = None,
+            deploy_gas: Optional[EtherValue] = None) -> ZethClient:
```
AntoineRondelet: Why not pass it in Wei directly?

pyClient/zeth/joinsplit.py
```diff
@@ -493,17 +668,16 @@ def _encrypt_note(out_note: ZethNote, pub_key: EncryptionPublicKey) -> bytes:
 def receive_notes(
         addrs_and_ciphertexts: List[Tuple[int, bytes]],
         sender_k_pk: EncryptionPublicKey,
-        receiver_k_sk: EncryptionSecretKey) -> Iterable[Tuple[int, ZethNote]]:
+        receiver_k_sk: EncryptionSecretKey) -> Iterator[Tuple[int, ZethNote]]:
     """
     Given the receivers secret key, and the event data from a transaction
     (encrypted notes), decrypt any that are intended for the receiver.
     """
```
AntoineRondelet: Would you mind dropping a line to mention the structure of a note description returned by this function @dtebbs (basically what you said here https://github.com/clearmatics/zeth/pull/130#discussion_r359446403, a note desc being a note along with its address in the mktree)

pyClient/README.md
```diff
@@ -45,11 +55,226 @@ and fails to fetch it (or fails to fetch the right version).
...
+    --in 0 \                              # "b1a2feaf: value=200, addr=0"
+    --out d77f...0e00:cc7c....7f76,120 \  # 120 to this addr
+    --out 3a43...fd3b:9fc8....b838,180    # 180 to this addr
+```
+
+### Withdrawing funds from the mixer
+
+Specify the note(s) to be withdrawn, and the total value as public output:
+```console
+$ zeth mix --in eafe5f84 --vout 100
```
AntoineRondelet: Does the command crash if `--vout 100` is =/= from the value of the given note?
Say, what if I run:
```
zeth mix --in b1a2feaf --vout 100
```

Does it automatically create another note a value 100 belinging to me, and withdraws 100, or does it crash?

pyClient/README.md
```diff
@@ -45,11 +55,226 @@ and fails to fetch it (or fails to fetch the right version).
...
+
+### A note on the `zeth mix` command
+
+As explained above, the `zeth mix` command can be used to deposit funds on the
+mixer, transfer notes, and withdraw funds from the mixer. A single command can
+perform all of these in one transaction, which greatly improves the privacy
+level provided by Zeth. In fact, no exact information about the meaning of a
+transaction is ever leaked to the an observant attacker.
+
+Here are a few examples of complex payments allowed by `zeth mix`:
```
AntoineRondelet: `are a few examples` -> is  an example (sorry I think I made this typo as I intended to write several examples but ended up writing a single one).
Note that we can add another example where we do a "transfer-and-withdraw", ie: something like:
```
 zeth mix \
    --in eafe5f84 \                      # "eafe5f84: value=100, addr=2"
    --out d77f...0e00:cc7c....7f76,98.5 \ # 98.5 to this address (e.g. Bob)
    --vout 1.5
```

pyClient/README.md
```diff
@@ -45,11 +55,226 @@ and fails to fetch it (or fails to fetch the right version).
...
+
+$ zeth notes
```
AntoineRondelet: Should be `zeth ls-notes` since the command has been renamed

pyClient/README.md
```diff
@@ -45,11 +55,226 @@ and fails to fetch it (or fails to fetch the right version).
...
+## Environment
+
+Depending on the operation being performed, the `zeth` client must:
+- interact with an Ethereum RPC host,
+- interact with the deployed Zeth contracts,
+- request proofs and proof verification keys from `prover_server`, and
+- access secret and public data for the current user
+
+Since the primary and auxiliary inputs are generally very long, it is difficult
+to pass as arguments to the zeth commands. Such data is stored in files with
```
AntoineRondelet: to pass them

AntoineRondelet: Thus, such data [...]

README.md
```diff
@@ -80,22 +92,17 @@ npm install
...
 cd pyClient
-# Follow the few steps described in the README of the python stub
 ```
 
-### Use the pyClient
-
-This Proof of Concept comes with some minimal building blocks to integrate Zeth with your applications.
-You can use the [python client](pyClient/README.md) to interact with the proving service and request proofs on a given `(instance, witness)` pair.
-
-## Secure Multi Party Computation for the Groth16 SRS generation
```
AntoineRondelet: Please add this title back otherwise the mpc section gets added to the section about the python client

pyClient/README.md
AntoineRondelet: Quickly checked. This would crash: https://github.com/clearmatics/zeth/blob/pyclient-cli/pyClient/commands/zeth_mix.py#L60-L61

pyClient/zeth/joinsplit.py
dtebbs: `EthereValue` allows the caller to specify the value in any units, and handles the conversion from srting (e.g. from command line params) and ints in code.

pyClient/README.md
dtebbs: Hopefully, it outputs an error message :) but yes, it fails upfront.
We could add some support for change etc, btu for now I just enforce that everything is explicit and the values balance correctly.

pyClient/zeth/joinsplit.py
dtebbs: Added a description of the return values.

pyClient/README.md
dtebbs: Added this second example.  Thanks.

pyClient/README.md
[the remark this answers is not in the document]
dtebbs: @AntoineRondelet I've added a bit more to the README to try and make this clearer, including a directory structure based on your comment here.  I'm not sure if it's too much, but I thought it better to over-explain rather than leave room for confusion.  Let me know what you think.  It might make sense to reword some of it.  Thanks.

pyClient/zeth/wallet.py
```diff
@@ -6,40 +6,193 @@
...
+        return ZethNoteDescription(
+            note=joinsplit.zeth_note_from_json_dict(json_dict["note"]),
+            address=int(json_dict["address"]),
+            commitment=bytes.fromhex(json_dict["commitment"]))
+
+
 class Wallet:
-    def __init__(self, username: str, wallet_dir: str, k_sk_receiver: PrivateKey):
+    """
+    Very simple class to track a list of notes owned by a Zeth user. Note this
```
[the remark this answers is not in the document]
dtebbs: chaged

pyClient/README.md
dtebbs: Fixed.

pyClient/README.md
dtebbs: Updated

README.md
dtebbs: Oops.  Must have been accidentally removed somewhere.  Replaced.

pyClient/README.md
```diff
@@ -45,11 +55,265 @@ and fails to fetch it (or fails to fetch the right version).
...
+Depending on the operation being performed, the `zeth` client must:
+- interact with an Ethereum RPC host,
+- interact with the deployed Zeth contracts,
+- request proofs and proof verification keys from `prover_server`, and
+- access secret and public data for the current user
+
+Quite a lot of information must be given in order for the client to do this,
+and the primary and auxiliary inputs to a Zeth operation are generally very
+long. It can therefore be difficult to pass this information to the zeth
+commands as as command-line arguments. Thus, such data is stored in files with
```
AntoineRondelet: as as -> as

pyClient/README.md
```diff
@@ -45,11 +55,265 @@ and fails to fetch it (or fails to fetch the right version).
...
+  the zeth contract. This file is created by the deployment step below and
+  should be distributed to each client that will use this instance.
+- `zeth-address.json` and `zeth-address.json.pub` hold the secret and public
+  parts of a ZethAddress. These can be generated with the `zeth gen-address`
+  command. `zeth-address.json.pub` holds the public address which can be shared
+  with other users, allowing them to privately transfer funds to this client.
+  The secret `zeth-address.json` should **not** be shared.
+
+Note that by default the `zeth` command will also create a `notes`
+subdireectory to contain the set of notes owned by this user. These are also
```
AntoineRondelet: subdirectory

pyClient/README.md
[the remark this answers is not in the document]
AntoineRondelet: Thanks @dtebbs, just had a look, we can keep it that way for now. I'm sure there is place for a better management of the configuration but it's good to see that we can easily support mixing multiple assets by just following some conventions on how files are organized, and running the commands in the right directory. That's more than enough for now.

pyClient/README.md
dtebbs: fixed.  thanks.

pyClient/README.md
dtebbs: fixed.  thanks.

zeth-contracts/contracts/BaseMerkleTree.sol
```diff
@@ -70,4 +70,13 @@ contract BaseMerkleTree {
     // Returns the array of leaves of the merkle tree
     return tmpLeaves;
   }
+
+  // Retrieve a single tree entry.  We expect that clients will generally
+  // decrypt a very small fraction of new entris that appear in the tree, which
+  // represents an even smaller fraction of the total tree, so in general it is
+  // very inefficient for clients to download the entire tree.
+  function getLeaf(uint leaf_address) public view returns (bytes32) {
```
AntoineRondelet: Where do you use this @dtebbs ?

That quite a dangerous operation to carry out on the tree as it can leak deadly metadata about the commitments you are accessing in the tree!

AntoineRondelet: Manipulating the tree in its entirety is less efficient, not to say very inefficient, but it prevents several types of leakages since you never really know which entries "interest" the user

pyClient/zeth/wallet.py
```diff
@@ -6,40 +6,186 @@
...
+            return ZethNoteDescription.from_json(note_f.read())
+
+    def _check_note(
+            self, addr: int, note: ZethNote) -> Optional[ZethNoteDescription]:
+        """
+        Recalculate the note commitment that should have been stored in the
+        Merkle tree, and check that the commitment is at the correct address.
+        """
+        cm = joinsplit.compute_commitment(note)
+        mk_leaf = get_merkle_leaf(self.mixer_instance, addr)
```
AntoineRondelet: This is a big information leakage @dtebbs

pyClient/zeth/wallet.py
```diff
@@ -6,40 +6,186 @@
...
         new_notes = []
-        for addr, note in new_notes_iter:
-            print(
-                f"[INFO] {self.username} received payment: {note} (addr: {addr})")
-            self._write_note(note)
-            new_notes.append((addr, note))
+        addr_note_iter = joinsplit.receive_notes(
+            encrypted_notes, k_pk_sender, self.k_sk_receiver)
+        for addr, note in addr_note_iter:
+            note_desc = self._check_note(addr, note)
```
AntoineRondelet: @dtebbs I think this code is leaking a ton of information about the recipient since it calls the `_check_note` function which itself call the `get_merkle_leaf` function if the decryption is successful (ie: if you are the intended recipient of a payment).
Can you please confirm (or challenge if I misunderstand the code) ?

AntoineRondelet:
> > That quite a dangerous operation ...
> 
> I see .. So, we can get all leaves and check, or ome random subset, or just not perform this check. Sounds like you are suggesting that we get all leaves?

So by looking at the client code; here is what seems to happen.
1. The client reads the logs of the Mixer after `mix` has been called by a party on the network
2. The client wants to check if among the newly generated notes some have been encrypted under the key in the client config (ie. check if "we are the recipient"). As such the client decrypts the ciphertexts emitted
3. **If the decryption is successful**, then we check that the recovered plaintexts constitute a valid commitment opening of the commitments appended in the tree (https://github.com/clearmatics/zeth/blob/pyclient-cli/pyClient/zeth/wallet.py#L89-L95). The check process "pulls" the commitment at a given address from the Merkle tree on-chain (this is a smart contract call to the Merkle tree and so is a transaction from the recipient's address which gets broadcasted on the network: https://github.com/clearmatics/zeth/blob/dfb3e1f3daeaf56a7639464cda13d9354c23708a/pyClient/zeth/wallet.py#L130 and https://github.com/clearmatics/zeth/blob/dfb3e1f3daeaf56a7639464cda13d9354c23708a/pyClient/zeth/contracts.py#L454)
4. **Else (if the decryption is not successful)**, the check does not happen and so no smart contract call is made to fetch the commitment and verify the validity of the payment.

Looking at the above; we can see that the block:
```
if [decryption is sucessful] then:
    call_smart_contract(MKTree.get_merkle_leaf)
else:
     return []
```
leaks the recipient of a payment.

To deal with this situation, we need to make sure that both events (decryption is a success or a failure) **are indistinguishable** for other network members. As such, we can either:
- Modify the code to fetch the leaf no matter what the result of the decryption is; or
- Add the commitments (which are necessary for the payment verification on the recipient's side) to the data being read in step 1. of the above process. In other words, let's emit the commitments along the ciphertexts. In that way, the procedure would become:

1. The client reads the logs of the Mixer after `mix` has been called by a party on the network (the commitments are also logged)
2. The client wants to check if among the newly generated notes some have been encrypted under the key in the client config (ie. check if "we are the recipient"). As such the client decrypts the ciphertexts emitted
3. **If the decryption is successful**, then we check that the recovered plaintexts constitute a valid commitment opening of the commitments appended in the tree. This check is a **local** check.
4. **Else (if the decryption is not successful)**, the **local** check does not happen.

🚨 :warning: This solution is not perfect, especially if you consider adversaries with a physical access to the machine running the code, who may be able to carry out timing attacks to see whether you enter the "commitment check" block. However; it removes the need to call the MKTree smart contract as part of the "commitment check" process. To address such physical attacks we'd need to make sure that operations are constant time; and so we'd need to reshape the code to avoid having this `if/else` block that could leak (time and power) information about the computation being ran.

The modifs needed would be:

1. On the contract:
- Replace the `LogAddress` event by another event `LogCommitment`here :https://github.com/clearmatics/zeth/blob/master/zeth-contracts/contracts/BaseMixer.sol#L75-L80
- Emit the commitments: https://github.com/clearmatics/zeth/blob/master/zeth-contracts/contracts/BaseMixer.sol#L162

2. On the client:
- Read the LogCommitment events: https://github.com/clearmatics/zeth/blob/master/pyClient/zeth/contracts.py#L264-L267
- Replace this line: https://github.com/clearmatics/zeth/blob/dfb3e1f3daeaf56a7639464cda13d9354c23708a/pyClient/zeth/wallet.py#L130 by the **local** recomputation of the commitment

AntoineRondelet:
As a side note:
There are other tricky things that would need to be considered for the client/wallet development in order to "enforce" user best practices and thus minimize missteps of users.
For eg. if a user U who has never called the Mixer contract wants to withdraw some funds (i.e. call the `zeth mix` command with the `--vout` flag) while the Merkle tree does not have much commitments; then, we can see that U has received funds from a sender which is in a small set of potential senders (a few commitments in the tree => a few potential senders). So there are heuristics that could be followed on the client to minimize these missteps (ie. missteps like: withdraw as soon as the payment is received, withdraw when the tree is almost empty and so on) and minimize their consequences. For instance, we could extend the config of the client to keep track of the size of the Merkle tree. Instead of calling a function on the Merkle tree contract to check the MKtree size whenever you want to do a payment; we know that all clients inspect the logs emitted by the Mixer (to check if "the user is the recipient"). As such, we can think about adding a counter to the client configuration that gets incremented every time the function https://github.com/clearmatics/zeth/blob/pyclient-cli/pyClient/zeth/wallet.py#L80 gets called. Keeping a local record of the size of the Merkle tree (you can also imagine that you may be able to keep some info about the msg.sender of the txs corresponding to Mix calls to have an estimate on the number of users of the Mixer; ie. get an estimate on the size of the anonymity set; and so on and so on) could help coming up with heuristics and local checks to warn the user (display a scaring warning on the UI, or in the terminal; ask to confirm the action and so on) when an action on the Mixer could put the user's privacy at risk.

dtebbs:
Regarding the info leakage, as discussed, note that the information is leaked to the connected ethereum node rather than to the whole ethereum network (although other nodes may be able to obtain this info indirectly).  But still an important leak.  I'll implement your proposal to emit the commitment, so the set of rpc calls is consistent whether or not it was the recipient of a note.

AntoineRondelet:
> Regarding the info leakage, as discussed, note that the information is leaked to the connected ethereum node rather than to the whole ethereum network (although other nodes may be able to obtain this info indirectly). But still an important leak. I'll implement your proposal to emit the commitment, so the set of rpc calls is consistent whether or not it was the recipient of a note.

Yes thanks for pointing that out @dtebbs, that's important. **The `get_merkle_leaf` function does not emit a transaction, but rather does a `call()`!** This still generates a RPC call (network traffic) depending on whether we are recipient or not, so should be corrected.

Some recent attacks have exploited "differences in behavior of a user’s wallet upon scanning a new transaction, depending on whether the user is the intended recipient of that transaction". See https://crypto.stanford.edu/timings/paper.pdf for a recent example of such type of attacks.

dtebbs:
I've pushed a commit that removes the `get_merkle_leaf` and uses commitments emitted from the contract.

dtebbs:
I've added some further state to the wallet.  For now, this just tracks the number of notes in the mixer, but this example could be extended to track things like distinct ethereum sender addresses, etc.

zeth-contracts/contracts/BaseMerkleTree.sol
dtebbs: This has now been removed (as discussed below)

pyClient/zeth/wallet.py
dtebbs: We no longer perform a query here (see discussion about commitment events)

pyClient/zeth/contracts.py
```diff
@@ -21,8 +21,8 @@
 
 Interface = Dict[str, Any]
 
-# List of address and cipher text pairs
-EncryptedNote = Tuple[int, bytes]
+# Address, commit and cipher text tuples
```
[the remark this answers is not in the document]
dtebbs: Fixed

pyClient/zeth/joinsplit.py
```diff
@@ -666,20 +666,21 @@ def _encrypt_note(out_note: ZethNote, pub_key: EncryptionPublicKey) -> bytes:
...
+) -> Iterator[Tuple[int, bytes, ZethNote]]:
     """
     Given the receivers secret key, and the event data from a transaction
     (encrypted notes), decrypt any that are intended for the receiver. Return
     tuples `(<address-in-merkle-tree>, ZethNote)`. Callers should record the
     address-in-merkle-tree along with ZethNote information, for convenience
     when spending the notes.
     """
-    for address, ciphertext in addrs_and_ciphertexts:
+    for address, commit, ciphertext in addrs_and_ciphertexts:
```
AntoineRondelet: Let's rename the `addrs_and_ciphertexts` to account for the fact that we now also get the commitment. Maybe something like `addrs_commits_ciphertexts`, or more generally `event _data`? Now, `addrs_and_ciphertexts` suggests that there are only addresses and ciphertexts tracked by this variable which is not the case anymore

dtebbs: Renamed to `event_data`.  Thansk.

AntoineRondelet: Thanks @dtebbs

PR #127, opened 2019-11-20
Further pyclient types and refactoring

dtebbs:
Preparation for exposing zeth functionality as CLI tools.
- Big refactor of tests into calls to a single "joinsplit()' operation.
- Stricter types.
- Some naming more consistent with Zet paper

The cli tools are not complete, but this is a good place to merge, to avoid branches diverging too much.

PR #122, opened 2019-11-15
Fixing non-malleability vulnerability

rrtoledo:
h_sig was not checked to be the hash of the signing key vk (and other parameters) which led to a transaction malleability attack.

PR linked to issue https://github.com/clearmatics/zeth/issues/120

Issue #120, opened 2019-11-14
h_sig independant to vk


rrtoledo:
In the non-malleability upgrade, we introduced a few variables including h_sig. h_sig is used to compute the message authentication tags (h_i) and the new notes identifiers (h_i). h_sig is computed as the hash of the signature verification key (vk), some randomness (randomSeed) and the old notes nullifiers ({nf}_old).

The vulnerability is that we do not check anywhere that h_sig is correctly computed, and even provide it to the contract (see "verifyTransaction in [Zerocash full paper](http://zerocash-project.org/media/pdf/zerocash-extended-20140518.pdf) Figure 2 p21). An adversary could thus take the primary inputs (including h_sig) and the proof and create a new transaction with them (by faking the ciphertexts and creating a new verification key/signature).

To correct the vulnerability we need to force the contract to check the correctness of h_sig. We could,
- remove h_sig from the "inputs" variable and have the contract recompute them. This would be cheaper in term of variable but we would need extra computation to reassemble the primary inputs.
- check that h_sig = Hash( randomSeed,  vk, {nf_{old}}. We already reassemble the {nf_{old}} so this solution should cost us a hash.

AntoineRondelet:
Related to https://github.com/clearmatics/zeth/pull/43/files

AntoineRondelet:
Closing this issue since it has been addressed and merged

Issue #113, opened 2019-11-07
Further abstraction and  type information in pyClient


dtebbs:
- Replace values of the form `Tuple[str, str, int, bytes, bytes]` with meaningful types
- Abstract out many bits of functionality
  - details of creating commitments and signatures
  - snark-specific code
- ... other obvious refactorings ...

This will facilitate much of #8

AntoineRondelet:
Closing this issue as the corresponding PR has been merged

PR #132, opened 2019-12-13
Fixing non-malleability vulnerability (depends on #127)

dtebbs:
Copy of #122 applied on top of #127.
Thanks for the help @rrtoledo

### DIFF
--- a/pyClient/test/test_signing.py
+++ b/pyClient/test/test_signing.py
@@ -1,0 +1,33 @@
+from zeth import signing
+from hashlib import sha256
+from unittest import TestCase
+from os import urandom
+
+
+class TestSigning(TestCase):
+
+    def test_sign_verify(self) -> None:
+        """
+        Test the correct signing-verification flow:
+        verify(vk, sign(sk,m), m) = 1
+        """
+        m = sha256("clearmatics".encode()).digest()
+        keypair = signing.gen_signing_keypair()
+        sigma = signing.sign(keypair.sk, m)
+        self.assertTrue(signing.verify(keypair.vk, m, sigma))
+
+        keypair2 = signing.gen_signing_keypair()
+        self.assertFalse(signing.verify(keypair2.vk, m, sigma))
+
+    def test_sign_verify_random(self) -> None:
+        """
+        Test the correct signing-verification flow with random message:
+        verify(vk, sign(sk,m), m) = 1
+        """
+        m = urandom(32)
+        keypair = signing.gen_signing_keypair()
+        sigma = signing.sign(keypair.sk, m)
+        self.assertTrue(signing.verify(keypair.vk, m, sigma))
+
+        keypair2 = signing.gen_signing_keypair()
+        self.assertFalse(signing.verify(keypair2.vk, m, sigma))
--- a/pyClient/test_commands/scenario.py
+++ b/pyClient/test_commands/scenario.py
@@ -1,9 +1,12 @@
 import zeth.joinsplit as joinsplit
 import zeth.contracts as contracts
+from zeth.constants import ZETH_PRIME
+import zeth.signing as signing
 from zeth.utils import EtherValue, compute_merkle_path
 import test_commands.mock as mock
 import api.util_pb2 as util_pb2
 
+from os import urandom
 from web3 import Web3, HTTPProvider  # type: ignore
 from typing import List, Tuple
 
@@ -153,6 +156,48 @@ def charlie_double_withdraw(
     note1_value = joinsplit.to_zeth_units(EtherValue(CHARLIE_WITHDRAW_CHANGE_ETH))
     v_out = EtherValue(CHARLIE_WITHDRAW_ETH)
 
+    # ### ATTACK BLOCK
+    # Add malicious nullifiers: we reuse old nullifiers to double spend by
+    # adding $r$ to them so that they have the same value as before in Z_r,
+    # and so the zksnark verification passes, but have different values in
+    # {0;1}^256 so that they appear different to the contract.
+    # See: https://github.com/clearmatics/zeth/issues/38
+
+    attack_primary_input2: int = 0
+    attack_primary_input4: int = 0
+
+    def compute_h_sig_attack_nf(
+            nf0: bytes,
+            nf1: bytes,
+            sign_pk: joinsplit.JoinsplitSigVerificationKey) -> bytes:
+        # We disassemble the nfs to get the formatting of the primary inputs
+        input_nullifier0 = nf0.hex()
+        input_nullifier1 = nf1.hex()
+        nf0_rev = "{0:0256b}".format(int(input_nullifier0, 16))[::-1]
+        primary_input1_bits = nf0_rev[3:]
+        primary_input2_bits = nf0_rev[:3]
+        nf1_rev = "{0:0256b}".format(int(input_nullifier1, 16))[::-1]
+        primary_input3_bits = nf1_rev[3:]
+        primary_input4_bits = nf1_rev[:3]
+
+        # We perform the attack, recoding the modified public input values
+        nonlocal attack_primary_input2
+        nonlocal attack_primary_input4
+        attack_primary_input2 = int(primary_input2_bits, 2) + ZETH_PRIME
+        attack_primary_input4 = int(primary_input4_bits, 2) + ZETH_PRIME
+
+        # We reassemble the nfs
+        attack_primary_input2_bits = "{0:0256b}".format(attack_primary_input2)
+        attack_nf0_bits = \
+            attack_primary_input2_bits[256-3:] + primary_input1_bits
+        attack_nf0 = "{0:064x}".format(int(attack_nf0_bits[::-1], 2))
+        attack_primary_input4_bits = "{0:0256b}".format(attack_primary_input4)
+        attack_nf1_bits = \
+            attack_primary_input4_bits[256-3:] + primary_input3_bits
+        attack_nf1 = "{0:064x}".format(int(attack_nf1_bits[::-1], 2))
+        return joinsplit.compute_h_sig(
+            bytes.fromhex(attack_nf0), bytes.fromhex(attack_nf1), sign_pk)
+
     (output_note1, output_note2, proof_json, signing_keypair) = \
         zeth_client.get_proof_joinsplit_2_by_2(
             mk_root,
@@ -164,18 +209,20 @@ def charlie_double_withdraw(
             (charlie_apk, note1_value),  # recipient1
             (charlie_apk, 0),  # recipient2
             joinsplit.to_zeth_units(EtherValue(0)),  # v_in
-            joinsplit.to_zeth_units(v_out)  # v_out
-        )
+            joinsplit.to_zeth_units(v_out),  # v_out
+            compute_h_sig_attack_nf)
 
-    # ### ATTACK BLOCK
-    # Add malicious nullifiers (located at index 2 and 4 in the array of inputs)
-    # See: https://github.com/clearmatics/zeth/issues/38
-    r = 21888242871839275222246405745257275088548364400416034343698204186575808495617  # noqa
+    # Update the primary inputs to the modified nullifiers, since libsnark
+    # overwrites them with values in Z_p
+
+    assert attack_primary_input2 != 0
+    assert attack_primary_input4 != 0
+
     print("proof_json => ", proof_json)
     print("proof_json[inputs][2] => ", proof_json["inputs"][2])
     print("proof_json[inputs][4] => ", proof_json["inputs"][4])
-    proof_json["inputs"][2] = hex(int(proof_json["inputs"][2], 16) + r)
-    proof_json["inputs"][4] = hex(int(proof_json["inputs"][4], 16) + r)
+    proof_json["inputs"][2] = hex(attack_primary_input2)
+    proof_json["inputs"][4] = hex(attack_primary_input4)
     # ### ATTACK BLOCK
 
     # construct pk object from bytes
@@ -187,7 +234,7 @@ def charlie_double_withdraw(
         (output_note2, pk_charlie)])
 
     # Compute the joinSplit signature
-    joinsplit_sig = joinsplit.sign_mix_tx(
+    joinsplit_sig = joinsplit.joinsplit_sign(
         signing_keypair, sender_eph_pk, ciphertexts, proof_json)
 
     return zeth_client.mix(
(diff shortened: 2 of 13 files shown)


pyClient/zeth/joinsplit.py
```diff
@@ -601,46 +513,55 @@ def _compute_proof_hashes(proof_json: GenericProof) -> Tuple[bytes, bytes]:
...
             proof_elements.extend(proof_json[key])
     return (
-        sha256(encode_message_to_bytes(proof_elements)).digest(),
-        sha256(encode_pub_input_to_hash(proof_json["inputs"])).digest())
+        encode_message_to_bytes(proof_elements),
+        encode_message_to_bytes(proof_json["inputs"]))
 
 
 def sign_mix_tx(
+        signing_keypair: SigningKeyPair,  # Ephemeral signing key, tied to proof
```
AntoineRondelet: Should be a JoinsplitSigKeyPair, which itself is a wrapper around `SigningKeyPair`

```python
# JoinSplit Signature Keys definitions
JoinsplitSigKeyPair = signing.SchnorrKeyPair
JoinsplitSigPrivKey = signing.SchnorrSigningKey
JoinsplitSigPubKey = signing.SchnorrVerificationKey
```

Also, let's rename `sign_mix_tx` to `sign_joinsplit`, because we do not sign the transaction per say (ECDSA signed), here we only sign the tx_data corresponding to the Joinsplit description, so let's make it clear.

 (see: https://github.com/clearmatics/zeth/pull/122)

pyClient/zeth/joinsplit.py
```diff
@@ -601,46 +513,55 @@ def _compute_proof_hashes(proof_json: GenericProof) -> Tuple[bytes, bytes]:
...
         sender_eph_pk: EncryptionPublicKey,  # Ephemeral key used for encryption
         ciphertexts: List[bytes],  # Encyrpted output notes
         proof_json: GenericProof,  # Proof for the mix transaction
-        signing_keypair: SigningKeyPair  # Ephemeral signing key, tied to proof
 ) -> int:
+    """
+    Generate a Schnorr signature on the hash of the ciphertexts, proofs and
```
AntoineRondelet: `Schnorr signature` -> should be abstracted away from this file (ie: Schnorr should not be mentioned)

`joinsplit_sign` signs the joinsplit description with the `JoinSplitSigPrivKey`, by using the imported signing library. Let's keep the actual algo used and this function decoupled as much as possible

zeth-contracts/contracts/BaseMixer.sol
```diff
@@ -111,81 +111,42 @@ contract BaseMixer is MerkleTreeMiMC7, ERC223ReceivingContract {
...
 
-        // 2. We re-assemble the nullifiers (JSInputs)
+        // 2. We re-assemble the nullifiers (JSInputs) and check they were not already seen
+        bytes32[jsIn] memory nfs;
+        uint256[] memory digest_inputs = new uint[](2);
         for(uint i = 1; i < 1 + 2*jsIn; i += 2) {
             digest_inputs[0] = primary_inputs[i];
             digest_inputs[1] = primary_inputs[i+1];
-            bytes32 current_nullifier = Bytes.sha256_digest_from_field_elements(digest_inputs);
+            nfs[(i-1)/2] = Bytes.sha256_digest_from_field_elements(digest_inputs);
```
AntoineRondelet: `(i-1)/2` why use this as index?

pyClient/zeth/joinsplit.py
```diff
@@ -601,46 +513,55 @@ def _compute_proof_hashes(proof_json: GenericProof) -> Tuple[bytes, bytes]:
...
+    #   - proof elements
+    #   - public input elements
+    h = sha256()
+    h.update(encode_encryption_public_key(sender_eph_pk))
+    for ciphertext in ciphertexts:
+        h.update(ciphertext)
 
-    # Compute the joinSplit signature
-    return _sign(signing_keypair, hash_ciphers, proof_hash, pub_inputs_hash)
+    proof_hash, pub_inputs_hash = _compute_proof_hashes(proof_json)
```
rrtoledo: perhaps rename this function as we only encode now.

pyClient/zeth/joinsplit.py
```diff
@@ -601,46 +513,55 @@ def _compute_proof_hashes(proof_json: GenericProof) -> Tuple[bytes, bytes]:
...
-    blake_hash.update(nf0)
-    blake_hash.update(nf1)
-    for group_el in sign_pk:
-        for field_el in group_el:
-            blake_hash.update(field_el.n.to_bytes(32, byteorder="little"))
-    return blake_hash.digest()
+    h = sha256()
+    h.update(nf0)
+    h.update(nf1)
+    h.update(encode_vk_to_bytes(sign_pk))
```
rrtoledo: we should decide whether we call the verification key $vk$ or $pk$, here both notations are written which is a bit confusing.

pyClient/zeth/signing.py
```diff
@@ -1,39 +1,111 @@
...
 class SigningKeyPair:
     """
-    Key-pair for signing joinsplit data.
+    An OT-Schnorr signing and verification keypair.
     """
-    def __init__(self, sk: SigningSecretKey, pk: SigningPublicKey):
-        self.sk: SigningSecretKey = sk
-        self.pk: SigningPublicKey = pk
+    def __init__(self, x: FQ, y: FQ, x_g1: G1, y_g1: G1):
+        # We include in the signing key the verfication key
```
rrtoledo: type -> `verification key`

pyClient/zeth/joinsplit.py
dtebbs: Yes.  I'll switch to vk everywhere.  That's what the key structures use now.

pyClient/zeth/joinsplit.py
```diff
@@ -425,7 +327,10 @@ def joinsplit(
             outputs: List[Tuple[ZethAddressPub, EtherValue]],
             v_in: EtherValue,
             v_out: EtherValue,
-            tx_payment: EtherValue) -> contracts.MixResult:
+            tx_payment: EtherValue,
+            compute_h_sig_cb: Optional[ComputeHSigCB] = None,
+            modify_proof_json_cb: Optional[ModifyProofJsonCB] = None
```
AntoineRondelet: `modify_proof_json_cb: Optional[ModifyProofJsonCB] = None` do we actually use such callback?

pyClient/zeth/signing.py
```diff
@@ -1,39 +1,111 @@
...
+from hashlib import sha256
 from py_ecc import bn128 as ec
-from Crypto import Random
-from typing import Tuple, NewType
-
-
-FQ = ec.FQ
-G1 = Tuple[ec.FQ, ec.FQ]
+from zeth.utils import FQ, G1, encode_g1_to_bytes
+import zeth.constants as constants
```
rrtoledo: we only use `ZETH_PRIME`, it would be more efficient to do:
`from zeth.constants import ZETH_PRIME`

pyClient/test_commands/scenario.py
```diff
@@ -153,6 +155,45 @@ def charlie_double_withdraw(
...
     v_out = EtherValue(CHARLIE_WITHDRAW_ETH)
 
+    # ### ATTACK BLOCK
+    # Add malicious nullifiers: we reuse old nullifiers to double spend by
+    # adding $r$ to them so that they have the same value as before in Z_r,
+    # and so the zksnark verification passes, but have different values in
+    # {0;1}^256 so that they appear different to the contract.
+    # See: https://github.com/clearmatics/zeth/issues/38
+
+    def compute_h_sig_attach_nf(
```
AntoineRondelet: what does `_attach_nf` mean?

pyClient/zeth/joinsplit.py
```diff
@@ -184,8 +189,7 @@ def compute_nullifier(
     binary_ask = digest_to_binary_string(spending_authority_ask)
     first_252bits_ask = binary_ask[:252]
     left_leg_bin = "1110" + first_252bits_ask
-    left_leg = int(left_leg_bin, 2).to_bytes(32, byteorder='little')
-
+    left_leg = int(left_leg_bin, 2).to_bytes(32, byteorder='big')
```
rrtoledo: how did you come up in this change ? you checked if the value corresponded to the primary inputs ?

pyClient/test_commands/scenario.py
```diff
@@ -153,6 +155,45 @@ def charlie_double_withdraw(
...
+        nf0_rev = "{0:0256b}".format(int(input_nullifier0, 16))[::-1]
+        primary_input1_bits = nf0_rev[3:]
+        primary_input2_bits = nf0_rev[:3]
+        nf1_rev = "{0:0256b}".format(int(input_nullifier1, 16))[::-1]
+        primary_input3_bits = nf1_rev[3:]
+        primary_input4_bits = nf1_rev[:3]
+
+        # We perform the attack
+        attack_primary_input2 = int(primary_input2_bits, 2) + r
+        attack_primary_input4 = int(primary_input4_bits, 2) + r
```
AntoineRondelet: Why not pass the "mauled" nullifiers as arguments to the function directly?

In this function, you redefine the prime `r`, and do
```python
# We perform the attack
attack_primary_input2 = int(primary_input2_bits, 2) + r
attack_primary_input4 = int(primary_input4_bits, 2) + r
```

Which is done also a few lines later (outside of you callback)
```python
proof_json["inputs"][2] = hex(int(proof_json["inputs"][2], 16) + r)
proof_json["inputs"][4] = hex(int(proof_json["inputs"][4], 16) + r)
# ### ATTACK BLOCK
```

pyClient/test/test_signing.py
```diff
@@ -0,0 +1,33 @@
...
+
+        keypair2 = signing.gen_signing_keypair()
+        self.assertFalse(signing.verify(keypair2.vk, m, sigma))
+
+    def test_sign_verify_random(self) -> None:
+        """
+        Test the correct signing-verification flow with random message:
+        verify(vk, sign(sk,m), m) = 1
+        """
+        m = urandom(32)
```
rrtoledo: can you 
- change to a length strictly inferior than 32 (30 for instance)
- change `self.assertTrue(signing.verify(keypair.vk, m, sigma))` -> self.assertFalse(signing.verify(keypair.vk, m, sigma))
- add `self.assertTrue(signing.verify(keypair.vk, M, sigma))` with M being m encoded over 32 bytes ?

zeth-contracts/contracts/BaseMixer.sol
rrtoledo: This was to avoid using two different counters as we need one for digest_inputs and one for the nullifiers (or other variables).

We have 2 digest_inputs per nullifier (or other variable) so we need to divide $i$ by 2 to obtain the correct range for the nullifier index.
As the root is at index 0, i is odd ( 1, 3, 5... ) so we also need to remove 1 (not to rely on floor/ceil of the "/" operation).

pyClient/zeth/signing.py
rrtoledo: (I should have been more precised)

pyClient/zeth/joinsplit.py
rrtoledo: This should have been added for the "nullifier + r" attack. 
Because of the non malleability update, we compute `h_sig = sha256(vk || nf_0 || nf_1)`. For the "nullifier + r"  to pass, we need the `h_sig` check to pass. We thus need to update the value of `h_sig` to `h_sig^{rogue} = sha256(vk || nf_{0}^{rogue} || nf_1^{rogue})`
Similarly we need the proof check to pass, so we need to recompute the proof with `h_sig^{rogue}`.

pyClient/test_commands/scenario.py
rrtoledo: change to `compute_h_sig_attack_nf` please

pyClient/zeth/joinsplit.py
dtebbs: This came from a previous change I made (using bytes directly instead of converting back and forth from strings).  It only caused an error when the contract started checking the value as used in hsig.  I checked this against the original binary str -> int -> hex -> bytes code and realised it must end up big endian.
(As you noted in a discussion offline, the value get's recomputed by the prover server, which is why this error was not flagged until the h_sig check was made in the contract)

pyClient/zeth/joinsplit.py
AntoineRondelet: My question is about the callback: `modify_proof_json_cb: Optional[ModifyProofJsonCB]`, not the one for h_sig

zeth-contracts/contracts/BaseMixer.sol
rrtoledo: It should be more efficient to have a second counter actually (as we would have a C_variable_declaration + jsin*C_addition instead of jsin*(C_addition + C_division)  and C_var_declaration < C_division ).
 something like:

`
uint nf_count;\\
for(uint i = 1; i < 1 + 2*jsIn; i += 2) {\\
            digest_inputs[0] = primary_inputs[i];\\
            digest_inputs[1] = primary_inputs[i+1];\\
            bytes32 current_nullifier = Bytes.sha256_digest_from_field_elements(digest_inputs);\\
            nfs[nf_count] = Bytes.sha256_digest_from_field_elements(digest_inputs);\\
            nf_count = nf_count + 1;`

pyClient/zeth/joinsplit.py
rrtoledo: Sorry misread !

pyClient/test/test_signing.py
dtebbs: Sorry, can you explain this?  I don't follow.  It sounds like you are asking for another unit test.  Do you want to check that verification fails for a truncated message digest m[0:30]?  Or that signing fails if the message digest is too short?

rrtoledo: After thinking a bit more we dont need a unit test for that sorry. My thinking:

There is actually no problem per se.

It is just that on the client side we have a signature scheme which sign a message of any length and on the contract we have a verification function which verifies message of 32 bytes only.

An easy error we could have is that we sign a message which is 31 bytes (should not happen in our case because this is the output of sha256). On the contract side as this variable is cast as a bytes32 it is padded so the variable is different and the verification fails (because we hash the variable).
 
I just thought it would be good to have a reminder somewhere. A check on the message size on the signature algo would restrict it so I do not think it is a good idea. We don't need a unit test for that.

pyClient/zeth/joinsplit.py
dtebbs: Renamed

pyClient/zeth/joinsplit.py
dtebbs: Chaged

pyClient/zeth/signing.py
dtebbs: Updated

pyClient/zeth/joinsplit.py
dtebbs: renamed

dtebbs: Added the type wrappers back, renamed the sign functtion.

pyClient/zeth/joinsplit.py
dtebbs: Removed the reference to Schnorr

pyClient/zeth/joinsplit.py
dtebbs: removed

pyClient/test_commands/scenario.py
dtebbs: SHould have been `attack`.  Renamed now.

pyClient/test_commands/scenario.py
dtebbs: I've checked to use ZETH_PRIME and cached the public input values in the first instance.

The nullifier and h_sig are computed internally in the joinsplit module, hence the need for a callback instead of passing the computed values somewhere.  This avoids having extra fucntions in joinsplit which are purely for testing.

pyClient/zeth/signing.py
dtebbs: Reworded

pyClient/zeth/joinsplit.py
```diff
@@ -601,46 +512,55 @@ def _compute_proof_hashes(proof_json: GenericProof) -> Tuple[bytes, bytes]:
...
-def _compute_h_sig(
-        random_seed: bytes,
+def compute_h_sig(
         nf0: bytes,
         nf1: bytes,
-        sign_pk: SigningPublicKey) -> bytes:
+        sign_vk: JoinsplitSigVerificationKey) -> bytes:
     """
-    Compute h_sig = blake2s(randomSeed, nf0, nf1, joinSplitPubKey)
+    Compute h_sig = sha256(nf0 || nf1 || sign_pk)
```
AntoineRondelet: sign_vk

pyClient/test_commands/scenario.py
```diff
@@ -153,6 +156,48 @@ def charlie_double_withdraw(
...
+    # {0;1}^256 so that they appear different to the contract.
+    # See: https://github.com/clearmatics/zeth/issues/38
+
+    attack_primary_input2: int = 0
+    attack_primary_input4: int = 0
+
+    def compute_h_sig_attack_nf(
+            nf0: bytes,
+            nf1: bytes,
+            sign_pk: joinsplit.JoinsplitSigVerificationKey) -> bytes:
```
AntoineRondelet: sign_vk

pyClient/test_commands/scenario.py
```diff
@@ -153,6 +156,48 @@ def charlie_double_withdraw(
...
+        attack_primary_input2_bits = "{0:0256b}".format(attack_primary_input2)
+        attack_nf0_bits = \
+            attack_primary_input2_bits[256-3:] + primary_input1_bits
+        attack_nf0 = "{0:064x}".format(int(attack_nf0_bits[::-1], 2))
+        attack_primary_input4_bits = "{0:0256b}".format(attack_primary_input4)
+        attack_nf1_bits = \
+            attack_primary_input4_bits[256-3:] + primary_input3_bits
+        attack_nf1 = "{0:064x}".format(int(attack_nf1_bits[::-1], 2))
+        return joinsplit.compute_h_sig(
+            bytes.fromhex(attack_nf0), bytes.fromhex(attack_nf1), sign_pk)
```
AntoineRondelet: sign_vk

pyClient/zeth/joinsplit.py
dtebbs: should be fixed

pyClient/test_commands/scenario.py
dtebbs: should be fixed

pyClient/test_commands/scenario.py
dtebbs: fixed.  thanks.

zeth-contracts/contracts/BaseMixer.sol
AntoineRondelet: Can you please open a ticket for this one @rrtoledo, it might also be relevant to @dtebbs for some of the work carried out on the contracts.

rrtoledo: Already did !

AntoineRondelet: I assume it is #133 ? Please reference the related PR and issues together so that we can easily track the discussions down

AntoineRondelet:
Addressed: https://github.com/clearmatics/zeth/issues/120 on top of some refactor of the pyClient addressing: https://github.com/clearmatics/zeth/issues/113

Issue #140, opened 2019-12-18
Remove all PGHR13 files


AntoineRondelet:
Think about removing all support for PGHR13.

Now that the layer of abstractions are set to support various SNARKs in the backend, we don't need to support multiple of them necessarily (this is more potentially flawed code to maintain => Less is more)

Issue #103, opened 2019-10-31
Fix and update contract standalone tests


rrtoledo:
The standalone tests (in /zeth-contracts/test) do not work anymore.

This may be due to the migration scriptd being obsolete (e.g. the contracts were renamed and moved [here](https://github.com/clearmatics/zeth/blob/61a224a43eb7fe98e7a0ad93069fa83fa75b3830/zeth-contracts/migrations/2_deploy_contracts.js#L14) ) and some updates in the dependencies.

PR #141, opened 2019-12-18
Removed old contracts from the migration file and kept test contracts only

AntoineRondelet:
Address https://github.com/clearmatics/zeth/issues/103

Note: running `npm run test` in the `zeth-contracts` folder will throw an error "Stack too deep" when the `Pghr13Mixer.sol` contract is compiled.
This is not addressed in this PR. However, if one removes the `Pghr13Mixer.sol` contract (not tested in the tests) and run `npm run test`, the tests should now succeed.

Note: Fixing the "Stack too deep" error should be addressed in another PR. That being said, the support for PGHR13 may be dropped: https://github.com/clearmatics/zeth/issues/140 so this should be done (if necessary) once a decision to maintain or drop Pghr13 has been taken.


AntoineRondelet:
Freezed the PR for now, as it may make more sense to just get away from this JS tests and have them all in Python. Will unfreeze or close the PR later.

AntoineRondelet:
Deprecated by #367

Issue #144, opened 2020-01-08
Side-channel attack considerations


dtebbs:
Parts of the PoC implementation could / should be chaged to address some of these points:

- no query methods on contract
- clients keep merkle tree
- contract emits nullifiers
- periodic sync instead of on-demand
- (MAYBE) clients maintain set of "spent" notes
- (MAYBE) automatically add public vin/vout to create complex transactions
- (MAYBE) track more statistics and warn about anonymity set (based on heuristics)

AntoineRondelet:
Related: https://github.com/clearmatics/zeth/pull/130

AntoineRondelet:
Closing this issue since the corresponding PR has been merged

PR #145, opened 2020-01-15
Address some Client SCA issues

dtebbs:
Related to #144


dtebbs:
Currently WIP while the document is discussed and finalized.

pyClient/commands/utils.py
```diff
@@ -147,30 +149,62 @@ def open_wallet(
...
             mixer_instance = wallet.mixer_instance
             for mix_result in get_mix_results(
                     web3, mixer_instance, wallet_next_block, chain_block_number):
+                # For each result, write new commitments to the merkle tree, and
+                # then attempt to decrypt and validate notes intended for us.
+
+                for out_ev in mix_result.output_events:
+                    merkle_tree.set_entry(
+                        out_ev.commitment_address, out_ev.commitment)
+                new_merkle_root = mix_result.new_merkle_root
```
AntoineRondelet: Storing the entire Merkle Tree will be very expensive for the clients (2^{depth + 1} - 1 nodes to store, so if depth is set to 32, the storage requirements for the client will be very important). Instead of storing the full MK tree, the client is only interested in storing the nodes required to generate Merkle authentication paths for the commitments the client user "owns" (i.e.knows the opening for).
This may lead to various SCAs on the client itself, but won't be exploitable at the networking level which is what this PR seems to address

AntoineRondelet: Note: this does not need to be addressed on this PR, but is something to be aware of, and document for further iterations on the client.

dtebbs: Yes.  Noted for future work and documentations.

pyClient/commands/constants.py
```diff
@@ -20,3 +20,5 @@
 
 WALLET_DIR_DEFAULT = "./notes"
 WALLET_USERNAME = "zeth"
+
+MERKLE_TREE_FILE_DEFAULT = "./merkle_tree"
```
AntoineRondelet: Let's be consistent with the naming of the constants:
In the same file you have:
`INSTANCEFILE_DEFAULT = "zeth-instance.json"`, `ADDRESSFILE_DEFAULT = "zeth-address.json"` (`FILE` is concatenated directly without `_`), and now you declare `MERKLE_TREE_FILE_DEFAULT = "./merkle_tree"`, with a `_FILE`.

Let's stick to `_FILE` everywhere.

AntoineRondelet: Also, let's use file extensions in order to inform on how the data is stored and should be read (`.dat` can be used for generic data files)

pyClient/commands/zeth_ls_commits.py
```diff
@@ -15,12 +14,7 @@ def ls_commits(ctx: Any) -> None:
     """
     List all commitments in the joinsplit contract
     """
-    web3 = open_web3_from_ctx(ctx)
-    mixer_desc = load_mixer_description_from_ctx(ctx)
-    zeth_instance = mixer_desc.mixer.instantiate(web3)
-    null = bytes(32)
+    merkle_tree = open_merkle_tree(ctx)
```
AntoineRondelet: `open_merkle_tree` should receive a `Context`, so let's switch to  `Context` instead of `Any` in the signature of the `ls_commit` function

pyClient/commands/zeth_ls_notes.py
```diff
@@ -4,13 +4,16 @@
...
+from click import command, option, pass_context
 from typing import Any
 
 
 @command()
+@option("--balance", is_flag=True, help="Show total balance")
+@option("--spent", is_flag=True, help="Show spent notes")
 @pass_context
-def ls_notes(ctx: Any) -> None:
+def ls_notes(ctx: Any, balance: bool, spent: bool) -> None:
```
AntoineRondelet: Same here, `Any` should be `Context`

pyClient/test/test_merkle_tree.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+from zeth.merkle_tree import MerkleTree, PersistentMerkleTree, ZERO_ENTRY
+from os.path import exists, join
+from os import makedirs
+from shutil import rmtree
+from unittest import TestCase
+
+MERKLE_TREE_TEST_DIR = "_merkle_tests"
+MERKLE_TREE_TEST_SIZE = 16
```
AntoineRondelet: Let's rename this constant to be clearer. I imagine that by "size" you mean the nb of leaves and not the depth, right?

pyClient/commands/constants.py
dtebbs: Chaged to `_FILE_` everywhere, and used `.dat` for merkle tree data.

pyClient/commands/zeth_ls_commits.py
dtebbs: Used `Context` everywhere.

pyClient/commands/zeth_ls_notes.py
dtebbs: (fixed)

pyClient/test/test_merkle_tree.py
dtebbs: Renamed the constant.

pyClient/commands/zeth_ls_notes.py
AntoineRondelet: Thanks @dtebbs. I like the changes you have made to display the balance and the spent notes!

PR #147, opened 2020-01-16
Issue146 fix

### LINKED ISSUE #146: Contract compilation error with solc-v0.6.1
Currently if a user clones and executes Zeth from scratch will incur in a compilation error caused by `py-solcx-v0.1.1` compiling with `solc-v0.6.1` and deprecated flags. This happens if no compiler is available and, by default, the latest version is downloaded. More recent releases of `py-solcx` avoid this problems since they check the compiler version and filter unsupported options.  Updating `py-solcx` would solve this problem but it will introduce two different issues:

-  `solc-v.0.6.1` cannot be used in any case with our contracts since we have to stick to versions 0.5.x.

-  compiler installation folder used by `py-solcx` has changed and now `$HOME/.solcx` is used. So we are not able anymore to take everything into the virtual environment if we install the compiler by using the wrapper.

At the moment, a possible  solution is to update `py-solcx` and then force the installation of the right version of the compiler, either before the `make setup` of the python client or by adding the related installation command in the tests i.e. `install_solc('v0.5.x')`. The compiler has to be installed into the `$HOME/.solcx` folder since it's the place where `py-solcx` looks for. 
To keep everything into the virtual environment we should propose to the `py-solcx` mainteiners to support a custom installation folder for the compiler.

riemann89:
A brief summary follows:
- This fixes #146 issue by updating `py-solcx` version (and related dependencies) and forcing to install the correct version of the solidity compiler. I have moved all token contract functions (compile/deploy) into the `contracts.py` in order to run the installation of the compiler just once.

- I have also added also an instruction in `mypy` configuration file to ignore  `solcx` importing errors. The use of single `# type: ignore` instruction may result in a long line that make `flake8` complaining. Splitting the line makes `# type: ignore` not working anymore (see https://github.com/python/mypy/issues/500#issuecomment-86340780). There is a workaround but `flake8` will complain again. 
This is not strictly related to this issue, but would avoid problems in the future. We could evaluate if substitute all ignore instructions we have in the project.


pyClient/README.md
```diff
@@ -38,18 +35,7 @@ test_erc_token_mixing.py [ZKSNARK]
...
-$ wget https://github.com/ethereum/solidity/releases/download/[solc-version]/[solc-for-your-distribution] -O $ZETH/pyClient/zeth-devenv/lib/[python-version]/site-packages/solcx/bin/solc-[solc-version]
-$ chmod +x $ZETH/pyClient/zeth-devenv/lib/[python-version]/site-packages/solcx/bin/solc-[solc-version]
-```
+## Note on solc compiler installation
 
-To run this command you need to replace the solidity version (denoted by
-`[solc-version]`), the python version (denoted `[python-version]`), and binary
-file (denoted by `[solc-for-your-distribution]`) by your system specific pieces
-of information.
+Note that `make test` will automatically install solidity compiler in `$HOME$/.solc`
```
dtebbs: This makes it sounds as though one must run `make test` before using the library / commands.
Is the compiler not installed on-demand (as required)?  Maybe we can just say this?

pyClient/zeth/contracts.py
```diff
@@ -23,6 +23,9 @@
 
 Interface = Dict[str, Any]
 
+# Installing and setting the right version of the compiler to use
+install_solc_pragma(constants.SOLC_VERSION_PRAGMA)
```
dtebbs: Just one suggestion: If I've understood the setup correctly, the compiler will be installed whenever the `contract` module is included.  We could either execute `python -m zeth.contracts` as part of `make setup`, or (maybe better) we could do the install only when oen of the `compile` functions in this module is called.  In that way, clients that do not deploy the contract can avoid the install altogether.

riemann89: > we could do the install only when the compile functions in this module is called.

This solution probably is the best, but there are two annoying things:

- since we have many different `compile_*` primitives, if we add the install instruction multiple time we will see many times this log message, that we are unable to make it silent:

``
solc, the solidity compiler commandline interface                                                                                            
Version: 0.5.16+commit.9c3226ce.Linux.g++   
``

- repetitive instruction: 

`install_solc_pragma(constants.SOLC_VERSION_PRAGMA)`

repeated many times.

pyClient/README.md
riemann89: Yes, I'll modify it by removing `make test`.

pyClient/zeth/contracts.py
riemann89: What about we add 

`install_solc_pragma(SOLC_VERSION_PRAGMA)` 

into `setup.py` script?

dtebbs: Yeah, tbh either of those mechanisms seems fine.  Just a few comments.

> since we have many different compile_* primitives, if we add the install instruction multiple time we will see many times this log message, that we are unable to make it silent

This could be addressed if there is some kind of `get_installed_solc_versions`, so we can avoid unnecessary calls to `install_solc_pragma`.

But as part of the optimization work I switched to have just one contracat so one compile function (or maybe 2, including the token).  So practically, this message will only be seen once when deploying, and maybe once when the token is deployed.  (And I guess in general the token deployment will be separate from the zeth deployment).

> What about we add `install_solc_pragma(SOLC_VERSION_PRAGMA)` into setup.py script?

If it's difficult to do the above then installing from setup.py (or anywhere as part of `make setup`) is fine I think.

The main reason I'm focussing on making it on-demand is just that it's a shame to do the install on clients that do not require it (it installs dependencies and does a build on mac).  But this is not a huge issue, so not worth putting a lot of effort into at this stage.

I'm happy with whichever of these you prefer.

riemann89: Alright, to me make sense to separate the compiler installation from the call of compiling/deploy a contract and make it as part of the setup phase.

Adding this to the `make setup` seems working:

```
python -c "from solcx import install_solc_pragma; from zeth.constants import SOLC_VERSION_PRAGMA; install_solc_pragma(SOLC_VERSION_PRAGMA)"
```

I'll push it like these and then we can change it in the future if we find something better.

pyClient/README.md
```diff
@@ -38,18 +35,7 @@ test_erc_token_mixing.py [ZKSNARK]
...
-$ wget https://github.com/ethereum/solidity/releases/download/[solc-version]/[solc-for-your-distribution] -O $ZETH/pyClient/zeth-devenv/lib/[python-version]/site-packages/solcx/bin/solc-[solc-version]
-$ chmod +x $ZETH/pyClient/zeth-devenv/lib/[python-version]/site-packages/solcx/bin/solc-[solc-version]
-```
+## Note on solc compiler installation
 
-To run this command you need to replace the solidity version (denoted by
-`[solc-version]`), the python version (denoted `[python-version]`), and binary
-file (denoted by `[solc-for-your-distribution]`) by your system specific pieces
-of information.
+Note that `make setup` will automatically install solidity compiler in `$HOME$/.solc`
```
dtebbs:
```suggestion
Note that `make setup` will automatically install the solidity compiler in `$HOME/.solc`
```

pyClient/Makefile
```diff
@@ -4,6 +4,9 @@
 setup:
 	pip install --upgrade pip
 	pip install -e .
+	python -c "from solcx import install_solc_pragma; \
```
riemann89: Turns out that we have the following error in Travis:

`github ''API rate limit exceeded for 35.196.72.151. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)'`

I checked and 35.196.72.151 is a travis IP. The only way to remove this problem, that sporadically arises, is to avoid that these instructions are executed in the test. Any clue?

AntoineRondelet:
I think we need to update the python-client to use the standard json compilation with `compile_standard` instead of the legacy combined json compilation, with `compile_files`/`compile_source`, that way we can stick to the last version of solc

riemann89:
@AntoineRondelet at the moment I do not know if we can easily move from `^0.5.0 pragma` to `pragma ^0.6.0` in Zeth contracts without not trivial changes to the contracts (I'll have a look). To me would be better to have a temporarily fix for #146 and then open another issue in order to upgrade the contracts and allowing the use of the new major for the compiler.

pyClient/Makefile
dtebbs: Did you find a way to address this?
Let's make a `check_and_install_solc` function in contracts to check the installed version and only call install if necessary, to avoid this query.  Also, I think it makes sense to call that `check_and_install_solc` function from within the compile functions, so that we don't do this install unless it's genuinely required.

dtebbs:
> To me would be better to have a temporarily fix for #146 and then open another issue in order to upgrade the contracts and allowing the use of the new major for the compiler.

I've created #149 to deal with moving to the 0.6.x compiler.

pyClient/Makefile
riemann89: It seems that `py-solcx` already checks if the compiler is installed in the system, so it does it only if it's required.  A solution could be to generate a API token for the script ( https://github.blog/2013-05-16-personal-api-tokens/) and add it to the repo. Just realized it was suggested in the pysolcx codebase: https://github.com/iamdefinitelyahuman/py-solc-x/blob/32dfb7b8c378b61d143d44e1bf49da25c42fd151/solcx/install.py#L164

AntoineRondelet:
Alright, thanks guys, we can keep it that way then, i.e.~have a quick fix to fetch the latest version of the 0.5.x solc compiler and then switch to 0.6.x in a second time.

pyClient/Makefile
riemann89: The github api call has been removed, so we should be fine now. Travis run without problems after the last commit. I'll close the conversion.

riemann89:
@dtebbs Everything modified as discussed off-line. It works properly.

dtebbs:
@riemann89 @AntoineRondelet 
Great.  Then I think it's ready to merge.

AntoineRondelet:
Alright, looks good to me, thanks for the fix guys.
Let's move to the latest version of the compiler when possible, and let's switch the method of compilation of the contracts to `compile_standard`

Issue #142, opened 2019-12-19
Make digest packing bit order match the EVM


dtebbs:
Currently, when digests are packed into multiple field elements, the bits are ordered so that bit0 of the digest has lowest order, and bit255 has highest.

For big-endian machines, the full set of 256 bits must be reversed.  For little-endian machines, the bits in each byte must be reveresed.  In particular, this wastes gas in the mixer contract (which hits the big-endian case above).

AntoineRondelet:
Closing this issue since the associated PR has been merged

PR #148, opened 2020-01-20
Reversing primary inputs bit order

rrtoledo:
This PR is to optimize the packing update by modifying the circuits to encode the primary inputs in big endian. This lead to a simplified re-assembly of the primary inputs on the contract and thus demand less gas.


dtebbs:
Addresses #142

PR #154, opened 2020-01-31
Benchmarks at depth 32

dtebbs:
Depends on #145.

- Perform benchmarks on larger merkle tree depths
- Address code issues discovered during this process


dtebbs:
- Addressed merkle tree efficiency
- Fix int overflow in circuit creation
- ...
- Measured costs for several depths
- Switched to depth 32 by default

```
     |-------+------------+--------------+---------+--------------+------------|
     | DEPTH | Deploy Gas | Proof Time 1 |   Gas 1 | Proof Time 2 | Gas Cost 1 |
     |-------+------------+--------------+---------+--------------+------------|
     |     8 |    3074532 |        14.14 |  883923 |        13.86 |     778481 |
     |    16 |    3391362 |        13.95 | 1208195 |        14.43 |     983137 |
     |    32 |    4024620 |        14.77 | 1857793 |        14.78 |    1392351 |
```

dtebbs:
Hit another code-related issue when using other depth sizes.  Should be fixed now, so removing the WIP.

AntoineRondelet:
Can you please change the target branch to `develop` @dtebbs?

dtebbs:
> Can you please change the target branch to develop @dtebbs?

I've moved the target branch and rebased.  Thanks.

pyClient/test_commands/test_merkle_tree_contract.py
```diff
@@ -0,0 +1,116 @@
...
+            print(f" {layer_hex}")
+        raise Exception(f"failed")
+
+
+# def assert_root(expect_root: bytes, root: bytes, msg: str) -> None:
+#     if root != expect_root:
+#         print(f"FAILED: {msg}")
+#         print(f"EXPECTED: {expect_root.hex()}")
+#         print(f"ACTUAL  : {root.hex()}")
+#         raise Exception(f"failed")
```
AntoineRondelet: Did you keep this comment on purpose @dtebbs or should it be removed?

pyClient/zeth/joinsplit.py
```diff
@@ -326,7 +327,7 @@ def get_dummy_input_and_address(
         value=ZERO_UNITS_HEX,
         rho=get_dummy_rho(),
         trap_r=trap_r_randomness())
-    dummy_note_address = 7
+    dummy_note_address = 2 ** constants.ZETH_MERKLE_TREE_DEPTH - 1
```
AntoineRondelet: Note that the dummy note address could be chosen arbitrarily, because if the note value is 0, then the recomputed merkle root in the tree is not compared to the actual merkle root (i.e. the merkle auth path is not checked). May be worth adding a comment to clarify that this address is picked arbitrarily

pyClient/zeth/joinsplit.py
```diff
@@ -515,6 +513,9 @@ def joinsplit(
         outputs_with_a_pk = \
             [(zeth_addr.a_pk, to_zeth_units(value))
              for (zeth_addr, value) in outputs]
+
+        timer = Timer.started()
```
AntoineRondelet: Let's drop a comment above this line to explain that this has been added to benchmark the proof generation on the prover - for clarity (although one could argue that a timer dropped in the code is rarely to do something else). Note also that the timer is started before the RPC call -  so there is some latency that should be taken into consideration in the measure.

pyClient/zeth/utils.py
```diff
@@ -23,11 +23,16 @@
 from typing import List, Tuple, Union, Any, cast
 
 
+WEB3_HTTP_PROVIDER_TIMEOUT = 60 * 60
```
AntoineRondelet: Why 60*60?
Let's add the units to the name -> `WEB3_HTTP_PROVIDER_TIMEOUT_SEC`?

src/circuits/notes/note.tcc
```diff
@@ -5,6 +5,8 @@
 // Content Taken and adapted from Zcash
 // https://github.com/zcash/zcash/blob/master/src/zcash/circuit/note.tcc
 
+#include "note.hpp"
```
AntoineRondelet: The tcc is already included in the hpp

zeth-contracts/contracts/MerkleTreeMiMC7.sol
```diff
@@ -7,29 +7,110 @@ pragma solidity ^0.5.0;
...
-contract MerkleTreeMiMC7 is BaseMerkleTree {
+// The Merkle tree implementation must trade-off complexity, storage,
+// initialization cost, and update & root computation cost.
+//
+// This implementation stores all leaves and nodes, skipping those that have not
+// been populated yet.  The final entry in each row stores the default value for
+// that row.
+contract MerkleTreeMiMC7 is BaseMerkleTree
+{
+    uint256 constant MASK_LS_BIT = uint(~1);
```
AntoineRondelet: What is the `~` here?

zeth-contracts/contracts/MerkleTreeMiMC7.sol
```diff
@@ -7,29 +7,110 @@ pragma solidity ^0.5.0;
 import "./BaseMerkleTree.sol";
 import "./MiMC7.sol";
 
-contract MerkleTreeMiMC7 is BaseMerkleTree {
+// The Merkle tree implementation must trade-off complexity, storage,
+// initialization cost, and update & root computation cost.
+//
+// This implementation stores all leaves and nodes, skipping those that have not
+// been populated yet.  The final entry in each row stores the default value for
```
AntoineRondelet: `each row` -> `each layer`

pyClient/commands/utils.py
```diff
@@ -185,8 +186,8 @@ def _do_sync() -> int:
                 # then attempt to decrypt and validate notes intended for us.
 
                 for out_ev in mix_result.output_events:
-                    merkle_tree.set_entry(
-                        out_ev.commitment_address, out_ev.commitment)
+                    merkle_tree.insert(out_ev.commitment)
```
AntoineRondelet: Seems like you no longer use the `commAddr` entry of the `LogCommitment` event. I'd be tempted to keep the "set_entry" function to insert a commitment at a given index in the tree; but if we remove the function (as you did in this PR); then is it worth keeping the `commAddr` entry in the event?

src/circuits/notes/note.tcc
dtebbs: Yes, technically it's redundant, but it allows the file to be opened in an IDE with all symbols in scope.

zeth-contracts/contracts/MerkleTreeMiMC7.sol
dtebbs: It's the not operator.  I unset the least-significant bit using this.

pyClient/commands/utils.py
dtebbs: Yeah, good point.  I guess it's been redundant ever since the client-side merkle tree was put in place. I looked quickly into removing it.  It shouldn't be diffcult, but it's not quite trivial since the wallet relies on it (as it doesn't have access to the Merkle tree at the moment).  So, it's a small refactor, but I'd lean towards creating an issue and dealing with it in a separate small PR.

dtebbs: Created issue #162

pyClient/zeth/utils.py
dtebbs: This was left form debugging.  Renamed and set to 1min which should be much more than enough.

pyClient/zeth/joinsplit.py
dtebbs: Added a comment.

> Note also that the timer is started before the RPC call - so there is some latency that should be taken into consideration in the measure.

To be clear, this is what I meant by "round-trip time" (hopefully this is negligible compared to the current ~15s proof time).

pyClient/zeth/joinsplit.py
dtebbs: Added a comment, and switched to 0.

pyClient/test_commands/test_merkle_tree_contract.py
dtebbs: Removed.  Thanks.

zeth-contracts/contracts/MerkleTreeMiMC7.sol
dtebbs: Changed.

pyClient/zeth/joinsplit.py
AntoineRondelet: > To be clear, this is what I meant by "round-trip time"

Ah yes, indeed! Thanks for adding a comment

pyClient/commands/utils.py
AntoineRondelet: > I guess it's been redundant ever since the client-side merkle tree was put in place

Seems like it could still be relevant even with the "client-side Merkle tree", since you may want to be able to insert a leaf at a given index depending how you scan the chain. That being said, since the "sync" procedure scans all blocks sequentially, the index of the commitment indeed becomes redundant (each commitments are added to the tree to the "next index" - which is what you do here with the "insert" function)

AntoineRondelet: > I'd lean towards creating an issue and dealing with it in a separate small PR.

That works, thanks

pyClient/zeth/joinsplit.py
AntoineRondelet: Thanks

pyClient/zeth/utils.py
```diff
@@ -22,12 +22,19 @@
 from py_ecc import bn128 as ec
 from typing import List, Tuple, Union, Any, cast
 
+# Development Ethereum implementations can cause a timeout if the contract
```
AntoineRondelet: > Development Ethereum implementations can cause a timeout

What do you mean here @dtebbs?

src/circuits/notes/note.tcc
AntoineRondelet: Ah I see. We can keep it that way then if it is easier for you; the preprocessor conditional inclusion directives (#ifdef ...) should take care of avoiding cycling dependencies

zeth-contracts/contracts/MerkleTreeMiMC7.sol
AntoineRondelet: Oh I see.
I was confused by the operator precedence (see: https://solidity.readthedocs.io/en/v0.6.2/miscellaneous.html#order-of-precedence-of-operators)
What about writing `uint256 constant MASK_LS_BIT = ~uint256(1);` instead?

pyClient/commands/utils.py
dtebbs: > Seems like it could still be relevant even with the "client-side Merkle tree", since you may want to be able to insert a leaf at a given index depending how you scan the chain.

Just to clarify, by "redundant", I was only referring to the `commAddr` event.
Regarding the method to insert leaves at arbitrary indices, I'd favour keeping it out for now.  It adds a bit of complexity to the implementation and would be unused code.

zeth-contracts/contracts/MerkleTreeMiMC7.sol
dtebbs: changed

pyClient/zeth/utils.py
dtebbs: rephrased

zeth-contracts/contracts/MerkleTreeMiMC7.sol
AntoineRondelet: Thanks

pyClient/commands/utils.py
AntoineRondelet: Yes, that works

pyClient/zeth/utils.py
AntoineRondelet: Thanks

Issue #164, opened 2020-02-28
Use Github Actions


AntoineRondelet:
Switch Travis to Github Actions: https://github.com/features/actions

AntoineRondelet:
Closing this issue since the corresponding PR has been merged.

PR #167, opened 2020-03-11
GitHub actions

dtebbs:
Switch from travis to github actions.
Addresses #164.


dtebbs:
Seems to be working now, so removing WIP.

.github/workflows/main.yml
```diff
@@ -0,0 +1,66 @@
+# This is a basic workflow to help you get started with Actions
```
AntoineRondelet: If that from a template you have used? maybe we can remove this line

.github/workflows/main.yml
```diff
@@ -0,0 +1,66 @@
...
+
+name: CI
```
AntoineRondelet: What about `zeth-ci` instead?

.github/workflows/main.yml
```diff
@@ -0,0 +1,66 @@
...
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  push:
+  pull_request:
```
AntoineRondelet: Woudn't it be possible to run the ci also on pushes on develop by specifying `master` and `develop` here ? (https://help.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#filtering-for-specific-branches-tags-and-paths)

.github/workflows/main.yml
dtebbs: Removed the comments.

.github/workflows/main.yml
dtebbs: Changed to `zeth-ci`

.github/workflows/main.yml
dtebbs: As far as I know, this *will* run the ci on pushes to master and develop (once this is merged).  It should work on any branch.  If we add a filter then I think it *only* runs on those branches.
Let me know if you think I've missed something.

AntoineRondelet: Ah right. Thanks

AntoineRondelet:
Thanks @dtebbs LGTM

PR #183, opened 2020-03-29
[WIP] Formatting primary inputs

AntoineRondelet:


AntoineRondelet:
The fact that I removed the surrounding quotes around all the hexadecimal strings (i.e. `"["0xa...f", "0xf...f"]"` was changed to `"[0xa...f, 0xf...f]"`) breaks some of the code on the contracts deployment side. I'll switch back.

PR #184, opened 2020-03-30
Pyclient packaging and factoring for reuse

dtebbs:
Some small changes that help when reusing the pyclient code in external projects.
Some headline changes:
- Mixer events as a single event call (simplify parsing by clients)
- `Wallet` class owns `MerkleTree`
- Remove the "Commit Address" from mixer event
- Basic packaging changes
- Factor out the Mix params as a serializable class

PR #189, opened 2020-04-08
Pyclient contract abstraction [depends on #184]

dtebbs:
Changes related to contract deployment, invocation etc.
Expose some useful utility functions in client code, on contracts and for testing.

### DIFF
--- a/pyClient/commands/zeth_deploy.py
+++ b/pyClient/commands/zeth_deploy.py
@@ -6,7 +6,6 @@ from commands.constants import INSTANCE_FILE_DEFAULT
 from commands.utils import \
     open_web3_from_ctx, get_erc20_instance_description, load_eth_address, \
     write_mixer_description, MixerDescription
-from zeth.contracts import InstanceDescription
 from zeth.mixer_client import MixerClient
 from zeth.utils import EtherValue
 from click import Context, command, option, pass_context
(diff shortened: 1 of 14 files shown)


pyClient/commands/zeth_deploy.py
```diff
@@ -43,14 +42,12 @@ def deploy(
     token_instance_desc = get_erc20_instance_description(token_address) \
         if token_address else None
 
-    zeth_client = MixerClient.deploy(
+    _zeth_client, mixer_instance_desc = MixerClient.deploy(
```
AntoineRondelet, line 45: What is  the leading  `_` in `_name_of_var` supposed to represent? (put otherwise, why did you rename `zeth_client` to `_zeth_client`?)

pyClient/test_commands/test_ether_mixing.py
```diff
@@ -50,7 +50,7 @@ def main() -> None:
 
     # Deploy Zeth contracts
     tree_depth = zeth.constants.ZETH_MERKLE_TREE_DEPTH
-    zeth_client = MixerClient.deploy(
+    zeth_client, _contract_desc = MixerClient.deploy(
```
AntoineRondelet, line 53: Same question as above: Does the leading "_" have a meaning here?

pyClient/commands/zeth_deploy.py
dtebbs: For variables, the leading `_` marks the variable as unused preventing the linting tools from complaining.

AntoineRondelet: Oh alright I see. Isn't it possible to replace these unused variables by a `_` then? (i.e. `_, mixer_instance_desc = MixerClient.deploy(`

dtebbs: AFAIK `_` is not treated as special, as it is in some languages, (except where it is used for "last value" in the REPL, IIRC) so it get's a value and type like any other variable. Hence it usually creates type conflicts if used multiple times per scope.  So I think it's possible, but not necessarily everywhere, and in Python it doesn't seem to have an advantage over the `_` prefixed version (which also includes information about the unused variable).  At least, that's my understanding.

AntoineRondelet: I see - it's not like in Go...
I just read that a way to get around that is to select the outputs of a function by doing something like `desired_output1, desired_output2 = function(args..)[indices_of_desired_outputs]`. Haven't tried it though...

dtebbs: Interesting. I hadn't seen that approach before.

PR #197, opened 2020-04-16
Switched to Alpine pkg manager in CI script

AntoineRondelet:


AntoineRondelet:
Opened https://github.com/clearmatics/zeth/issues/198 to address compilation warnings properly in a second time.

PR #193, opened 2020-04-09
Add Ik-cca encryption scheme

riemann89:
This PR provides an ik-cca encryption scheme for Zeth (see #2)

PR #199, opened 2020-04-20
Consistency between encryption library and RFC [depends on #193]

dtebbs:
Verified that the code is consistent with the relevant RFC.  Added comments explaining this with links to code.


pyClient/zeth/encryption.py
```diff
@@ -19,6 +19,55 @@
...
+# appropriate standards as follows. (links refer to specific versions of external
+# libraries, to ensure that line numbers are correct, but the descriptions are
+# expected to hold for all versions.)
+
+# As described in [Bernstein06], private keys may be generated as 32 random
+# bytes with bits 0, 1 and 2 of the first byte cleared, bit 7 of the last byte
+# cleared, and bit 6 of the last byte set. This happens at key generation
+# time. See:
+#
+#   https://github.com/openssl/openssl/blob/be9d82bb35812ac65cd92316d1ae7c7c75efe9cf/crypto/ec/ecx_meth.c#L81
```
dtebbs, line 49: Note, the execution path through the code is very difficult to follow. To be sure, I've also experimentally verified that keys are generated with the expectd properties.

pyClient/zeth/encryption.py
```diff
@@ -18,8 +18,64 @@
...
+#  International Workshop on Public Key Cryptography, 2006,
+#  <https://cr.yp.to/ecdh/curve25519-20060209.pdf>
+#
+# \[LangleyN18]
+#  "Chacha20 and poly1305 for ietf protocols."
+#  Adam Langley and Yoav Nir,
+#  RFC 8439, 2018,
+#  <https://tools.ietf.org/html/rfc8439>
+
+# This implementation makes use of the `cryptography` library with OpenSSL
```
AntoineRondelet, line 45: Thanks for double checking on that @dtebbs 
Can we change this comment to a module-level docstring in order to add the explanations below to the module documentation?

dtebbs: Moved to docstring

dtebbs:
Also tidied some other things in that file (specifically some duplicated functionality). Let me know if you prefer that split into another PR.

pyClient/zeth/encryption.py
AntoineRondelet: Thanks

pyClient/zeth/encryption.py
```diff
@@ -217,3 +234,23 @@ def decrypt(
     message = decryptor.update(ct_sym)
 
     return message
+
+
+def _exchange(sk: EncryptionSecretKey, pk: EncryptionPublicKey) -> bytes:
```
AntoineRondelet, line 239: What does `_` mean? Does it mean "internal function"?

pyClient/zeth/encryption.py
```diff
@@ -4,6 +4,57 @@
...
 # SPDX-License-Identifier: LGPL-3.0+
 
+"""
+Encryption operations for Zeth notes. Supports an `encrypt` operation using
+receivers public key, and a `decrypt` operation using the corresponding private
+key. `decrypt` fails (except with negligible probability) if the ciphertext was
+encrypted with a different public key.
+
+This implementation makes use of the `cryptography` library with OpenSSL
+nnbackend. For the avoidance of doubt, the implementation adheres to the
```
AntoineRondelet, line 14: `nnbackend` -> backend

pyClient/zeth/encryption.py
AntoineRondelet: (Same question for the `_` prefix used for the constants in the tests)

dtebbs: Yes, exactly.  The linters will complain if another module tries to use it, and I think they also use this to check for unused functions.

pyClient/zeth/encryption.py
dtebbs: Thanks.  Fixed.

pyClient/zeth/encryption.py
AntoineRondelet: I see, thanks

PR #202, opened 2020-04-21
Build flag management

dtebbs:


dtebbs:
Overview:
- build all dependencies with their own compile flags
- mark dependencies as SYSTEM, so that warnings in dependencies are not reported in the zeth build
- build all zeth code with `-Wall -Wextra`
- enable `-Werror` on CI server

(Note, the rationale here is that during development it can be useful to allow some warnings, but warnings should be caught in pushed code.  We can moify this if there are differing opinions.)

libzeth/test/powersoftau_test.cpp
```diff
@@ -15,10 +16,13 @@ using G1 = libff::G1<ppT>;
 using G2 = libff::G2<ppT>;
 using namespace libsnark;
 using namespace libzeth;
+namespace fs = boost::filesystem;
 
 namespace
 {
 
+fs::path g_testdata_dir = fs::path("..") / "testdata";
```
AntoineRondelet, line 24: Wouldn't it be possible to use an environment variable here? Otherwise we need to assume that the tests are written from a specific location (else, we end up writing `testdata` in the user's filesystem)

AntoineRondelet: Oh, just realized that you overwrote this global variable before running the tests..

Issue #203, opened 2020-04-22
Use of constants in client


dtebbs:
There are several places (notably in `mixer_client.py`, but probably in other places) where constants such as `APK_LENGTH` and `RHO_LENGTH` could be used when generating and validating data, replacing some hard-coded values.

AntoineRondelet:
Closing this issue has the corresponding PR has been merged

PR #204, opened 2020-04-23
Pyclient constant use

dtebbs:
- Consistent naming of constants, and removal of duplicates
- make some internal constants (implementation details) private
- replace some hard-coded sizes and offsets with constants


pyClient/zeth/constants.py
```diff
@@ -4,8 +4,27 @@
 #
 # SPDX-License-Identifier: LGPL-3.0+
 
+"""
+Constants used by zeth.  By convention lengths are given in bits as
+`*_LENGTH` and the corresponding `*_LENGTH_BYTES` variable holds the size in
+bytes (where this is meaningful.
```
AntoineRondelet, line 10: missing `)`

pyClient/zeth/constants.py
```diff
@@ -66,45 +97,22 @@
...
-MAC_KEY_LENGTH: int = 256
-TAG_LENGTH: int = 128
-KEY_MATERIAL: int = SYM_KEY_LENGTH + MAC_KEY_LENGTH
-SYM_NONCE_LENGTH: int = 128
-NOTE_LENGTH: int = APK_LENGTH + NOTE_VALUE_LENGTH + RHO_LENGTH + TRAPR_LENGTH
-ENCRYPTED_NOTE_LENGTH: int = EC_PUBLIC_KEY_LENGTH + NOTE_LENGTH + TAG_LENGTH
-
-# Encryption constants values
-SYM_NONCE_VALUE: int = 0
+COMMITMENT_VALUE_PADDING: bytes = bytes(int(192/8))
```
AntoineRondelet, line 118: Let's add a comment here. Using numbers (without labels) like here can be confusing. Here 192 comes from the fact that the value of the notes is encoded on 64 bits, and we need to pad it by 192 bits to get the other half of the input block for blake2s (which we use as commitment).

Do we actually use this constant somewhere else? I believe that we shouldn't because we do not call the compression function multiple times manually but rather switched to use the full hash function for the commitment, so we shouldn't need this padding anymore

AntoineRondelet:
Tackles: https://github.com/clearmatics/zeth/issues/203

pyClient/zeth/constants.py
dtebbs: Added `)`.  Thanks.

pyClient/zeth/constants.py
dtebbs: Yes, it seems unused now.  Removed.

PR #205, opened 2020-04-23
Fix for clang-10 build

dtebbs:
(Fixes recent CI build errors)


dtebbs:
Now contained in #202, to make that build pass.  I'll keep this open so we have the option of merging it if the build-flags PR is delayed.

PR #206, opened 2020-04-29
Refactor libsnark helpers [depends on #200]

dtebbs:
Rearranging files and the snark abstractions to parameterize by snark.
Part of a larger cleanup, this PR focussed on these two points and remaining clean up will happen in future PRs.


libzeth/circuits/binary_operation.hpp
```diff
@@ -119,6 +118,7 @@ class double_bit32_sum_eq_gadget : public libsnark::gadget<FieldT>
 };
 
 } // namespace libzeth
+
```
dtebbs, line 121: I've switched to having a space here consistently across all files.  It seems more consistent with the rest of the coding standards (we always leave an empty line after closing a function or namespace, and always have an empty line before includes), and matches the layout in libsnark.

AntoineRondelet: Sounds good @dtebbs (as long as we're consistent everywhere that's fine)

dtebbs:
Removed the WIP.  As above, this PR does not address refactoring of the serialization functions, or renaming.  It aims to decide on the file divisions / locations and to expse the snarks in a slightly more versatile way.  Further PRs will follow, but flag anything that you think makes sense to include here.

libzeth/core/ff_utils.hpp
```diff
@@ -0,0 +1,91 @@
+// Copyright (c) 2015-2020 Clearmatics Technologies Ltd
```
AntoineRondelet, line 1: Seems like you renamed `sciprlab_libs_utils` to `ff_utils`. These files do not manipulate finite field elements though, but also elliptic curve points, so the name doesn't really reflect the content

libzeth/snarks/default/default_api_handler.hpp
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) 2015-2020 Clearmatics Technologies Ltd
```
AntoineRondelet, line 1: Why not: `mv $ZETH/libzeth/snarks/default/* $ZETH/libzeth/snarks/`? (default is already part of the file names)

libzeth/core/joinsplit.hpp
```diff
@@ -2,11 +2,11 @@
 //
 // SPDX-License-Identifier: LGPL-3.0+
 
-#ifndef __ZETH_TYPES_JOINSPLIT_HPP__
-#define __ZETH_TYPES_JOINSPLIT_HPP__
+#ifndef __ZETH_CORE_JOINSPLIT_HPP__
```
AntoineRondelet, line 5: I'd be tempted to rename this file `joinsplit_input` as there is a mismatch between the name of the file and the class here

libzeth/core/utils.hpp
```diff
@@ -0,0 +1,83 @@
...
+/// This function throws an exception if called with an invalid hexadecimal
+/// string.
+std::vector<bool> hexadecimal_str_to_binary_vector(std::string str);
+
+/// Takes an hexadecimal digest and converts it into a binary vector.
+/// This function throws an exception if called with an invalid hexadecimal
+/// digest.
+std::vector<bool> hexadecimal_digest_to_binary_vector(std::string str);
+
+// Returns the little endian binary encoding of the integer x.
```
AntoineRondelet, line 42: ///

libzeth/core/utils.hpp
```diff
@@ -0,0 +1,83 @@
...
+///
+/// Throw if input is not well-formed. The type being checked should conform
+/// to the StructuredT interface above.
+template<typename StructuredT>
+void check_well_formed(const StructuredT &v, const char *name);
+
+// /// Throw if input is not well-formed. The type being checked should conform
+// /// to the StructuredT interface above.
+// template<typename StructuredT>
+// void check_well_formed_(const StructuredT &v, const char *name);
```
AntoineRondelet, line 72: Duplicate of the function above? Shall we can remove this comment?

libzeth/core/utils.hpp
```diff
@@ -0,0 +1,83 @@
...
+/// For some iterable container of objects comforming to StructuredT, throw if
+/// any entry is not well-formed.
+template<typename StructuredTs>
+bool container_is_well_formed(const StructuredTs &values);
+
+} // namespace libzeth
+
+#include "libzeth/core/utils.tcc"
+
+#endif // __ZETH_UTIL_HPP__
```
AntoineRondelet, line 83: `__ZETH_UTIL_HPP__` -> `__ZETH_UTILS_HPP__`

libzeth/core/extended_proof.hpp
```diff
@@ -2,36 +2,35 @@
 //
 // SPDX-License-Identifier: LGPL-3.0+
 
-#ifndef __ZETH_EXTENDED_PROOF_HPP__
-#define __ZETH_EXTENDED_PROOF_HPP__
+#ifndef __ZETH_TYPES_EXTENDED_PROOF_HPP__
```
AntoineRondelet, line 5: `__ZETH_TYPES_EXTENDED_PROOF_HPP__` this does not seem consistent with the rest. I think it should be `__ZETH_CORE_EXTENDED_PROOF_HPP__`

libzeth/core/multi_exp.tcc
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) 2015-2020 Clearmatics Technologies Ltd
+//
+// SPDX-License-Identifier: LGPL-3.0+
+
+#ifndef __ZETH_CORE_GROTH16_MULTI_EXP_TCC__
```
AntoineRondelet, line 5: `__ZETH_CORE_GROTH16_MULTI_EXP_TCC__`-> `__ZETH_CORE_MULTI_EXP_TCC__`

libzeth/core/utils.hpp
```diff
@@ -0,0 +1,83 @@
+// Copyright (c) 2015-2020 Clearmatics Technologies Ltd
+//
+// SPDX-License-Identifier: LGPL-3.0+
+
+#ifndef __ZETH_UTIL_HPP__
```
AntoineRondelet, line 5: `__ZETH_UTIL_HPP__` -> `__ZETH_CORE_UTILS_HPP__`

libzeth/core/ff_utils.hpp
dtebbs: `ff` was supposed to reflect `libff`, but I agree the name is not great.  I was trying to avoid using the name of dependent libraries or groups in these file names, so looking for something generic (hence `ff_utils` was also not satisfying).   I could separate it into `field_element_utils` and `group_element_utils`?  Or can you think of a better name?

libzeth/snarks/default/default_api_handler.hpp
AntoineRondelet: I realized that this was done for all files under `default`, `groth16` and `pghr13`. Why not rename the files from
```
default
|    |_ default_api_handler.hpp
|    |_ default_snark.hpp
|
|_ groth16
|      |_ groth16_api_handler.hpp
|      |_ groth16_snark.hpp
...
```

to 

```
default
|    |_ api_handler.hpp
|    |_ snark.hpp
|
|_ groth16
|      |_ api_handler.hpp
|      |_ snark.hpp
...
```

?

dtebbs: Just to be sure, are you suggesting this?
```
  libzeth/
    snarks/
      groth16_snark.*
      groth16_api_handler.*
      pghr13_snark.*
      pghr13_api_handler.*
      default_snark.*
      default_api_handler.*
```
I'm happy to go with that, if it's what you are referring to.

libzeth/circuits/circuit_constants.hpp
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) 2015-2020 Clearmatics Technologies Ltd
```
AntoineRondelet, line 1: These are "protocol constants" rather than "circuit constants". We also use them in the client etc (but since the client is in Python, we have a separate file). I would move these outside of the `circuits` folder and rename the file accordingly

libzeth/core/multi_exp.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+#ifndef __ZETH_CORE_MULTI_EXP_HPP__
+#define __ZETH_CORE_MULTI_EXP_HPP__
+
+#include "libzeth/core/include_libff.hpp"
+
+namespace libzeth
+{
+
+template<typename fieldT, typename groupT>
```
AntoineRondelet, line 13: fieldT -> `FieldT`
groupT -> `GroupT`

libzeth/core/bits.cpp
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) 2015-2020 Clearmatics Technologies Ltd
+//
+// SPDX-License-Identifier: LGPL-3.0+
+
+#include "libzeth/core/bits.hpp"
+
```
AntoineRondelet, line 6: Extraneous line?

libzeth/core/utils.hpp
dtebbs: I was going to mention this later, but it may be sensible now: are you happy to move to `#pragma once` for all include instead of these traditional include guards?  I could do that as part of this PR to fix all these I forgot to change.

libzeth/snarks/default/default_api_handler.hpp
dtebbs: Just seen the 2nd comment ... I wanted to avoid having multiple files with the same name (even in different dirs).  It can be confusing, and also doesn't match the pattern of naming files after the main type they correspond to.  (e.g. `groth16_snark` defines the `groth16_snark` type).

libzeth/mpc/groth16/powersoftau_utils.cpp
```diff
@@ -119,31 +119,28 @@ void write_powersoftau_fp2(
...
+void read_powersoftau_g1(std::istream &in, libff::alt_bn128_G1 &out)
 {
     uint8_t marker;
     in.read((char *)&marker, 1);
 
     switch (marker) {
     case 0x00:
         // zero
-        out = libff::G1<ppT>::zero();
+        out = libff::G1<srs_pot_pp>::zero();
```
AntoineRondelet, line 135: `libff::alt_bn128_G1 &out` is used as type of the 2nd function parameter, so is it really needed to use `libff::G1<srs_pot_pp>` in the function? We know that for the function to be sound `srs_pot_pp` will need to be `libff::alt_bn128` here

AntoineRondelet: Same applies for some files below

libzeth/snarks/groth16/groth16_api_handler.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include "libzeth/core/extended_proof.hpp"
+#include "libzeth/snarks/groth16/groth16_snark.hpp"
+
+#include <api/snark_messages.grpc.pb.h>
+
+namespace libzeth
+{
+
+/// Implemetation of API-related functions for the Groth16 snark.
```
AntoineRondelet, line 16: Implemetation -> Implementation

libzeth/snarks/pghr13/pghr13_api_handler.hpp
```diff
@@ -0,0 +1,48 @@
...
+#define __ZETH_SNARKS_PGHR13_PGHR13_API_HANDLER_HPP__
+
+#include "api/snark_messages.grpc.pb.h"
+#include "libzeth/snarks/pghr13/pghr13_snark.hpp"
+#include "libzeth/types/extended_proof.hpp"
+
+namespace libzeth
+{
+
+/// Implemetation of API-related functions for the Groth16 snark.
```
AntoineRondelet, line 15: Implemetation -> Implementation

AntoineRondelet: Groth16 -> PGHR13

libzeth/snarks/pghr13/pghr13_snark.tcc
```diff
@@ -0,0 +1,307 @@
...
+template<typename ppT>
+typename pghr13_snark<ppT>::KeypairT pghr13_snark<ppT>::generate_setup(
+    const libsnark::protoboard<libff::Fr<ppT>> &pb)
+{
+    // Generate verification and proving key (Trusted setup) from the R1CS
+    // (defined in the ZoKrates/wraplibsnark.cpp file) This function, basically
+    // reduces the R1CS into a QAP, and then encodes the QAP, along with a
+    // secret s and its set of powers, plus the alpha, beta, gamma, and the rest
+    // of the entries, in order to form the CRS (crs_f, shortcrs_f, as denoted
+    // in [GGPR12])
```
AntoineRondelet, line 24: The comment above can be removed IMO

libzeth/core/bits.cpp
dtebbs: I'm not sure why, but clang-format insists on putting thsi extra line in.  I'm not sure sure it happens here and nowhere else, but I can't find a way to avoid it.

libzeth/core/utils.hpp
AntoineRondelet: I saw that you used `#pragma once` is some files which breaks consistency with the rest of the code base. Using `#pragma once` is appealing: it is less code to write (and maintain - which makes this type of refactoring easier), and is nicer. However, I read some time ago that it wasn't yet included in the standard (although widely supported), and that it could be problematic in some edge cases. I don't know if the situation as changed but as of now I would stick to the include guards.

libzeth/snarks/default/default_api_handler.hpp
AntoineRondelet: Hum.. fair. I wouldn't have done it that way, but I am happy to stick to what you proposed.

libzeth/mpc/groth16/powersoftau_utils.cpp
dtebbs: Used the `libff::alt_bn128_??` types in these fundamental serialization functions.

libzeth/core/ff_utils.hpp
AntoineRondelet: The only name that comes to my mind is `ff_ec_utils.*`... I think the separation `field_element_utils` and `group_element_utils` may be better

libzeth/core/bits.cpp
dtebbs: TBH, I didn't look in detail (just tried removing it, but it always gets put back).  My guess is that clang-format groups the headers as "local header" "includes from local library" "external includes", with a space between each.

dtebbs: I guess clang doesn't apply this rule in `.tcc` files, as it doesn't consider them as "source" files.

libzeth/core/utils.hpp
dtebbs: I'll keep the traditional guards, then.

(AFAIK, the situation is that it's still not "standard", in the sense of being in the C++ standards, but supported by all major compilers for a while now).

AntoineRondelet: Right. Feel free to open a ticket to flag the possibility to switch to `#pragma once` in a second step. I am not fully opposed to that, but I would like to understand better the consequences of switching to that (especially the edge cases in which this can be problematic)

libzeth/circuits/circuit_constants.hpp
dtebbs: Yes, broader than the circuit. And related to the types in `circuit_types.hpp`, which could also be considered similar constants or configuration of the protocol. I've renamed it to `zeth_constants.hpp` and moved it to `libzeth`, which doesn't seem like the right place (since it's not relevant ot `core`, `snarks` or `serialization`), but it seems like overkill to create a `config` or `constants` directory for just this file, and if we wannted to put `circuit_typees.hpp` in there, `config` would have to depend on `circuits`, creating a circuilar dependency.

One option would be to have a `gadgets` dir with just the gadgets for the various commitments, annd then a `zeth` dir (or some other name, broader than circuits) which could contains these constants, the choices for Hash, HashTree, etc and the final `jonisplit` gadget.

For now, I'll keep it in the `libzeth`, but let me know if you have a better ideaa or a strong opinion about it.

AntoineRondelet: > For now, I'll keep it in the libzeth, but let me know if you have a better ideaa or a strong opinion about it.

Yes I'd just move it up the file hierarchy and rename accordingly for now

libzeth/core/bits.cpp
AntoineRondelet: > I guess clang doesn't apply this rule in .tcc files, as it doesn't consider them as "source" files.

Not sure I get the point. This file is a `cpp` file, not a `tcc` file, right? And both includes are `libzeth/core/*` so they should both be "local headers", no?

dtebbs: > Not sure I get the point. This file is a cpp file, not a tcc file, right? And both includes are libzeth/core/* so they should both be "local headers", no?

Since `.tcc` is not really a well known file extension, I assume clang-format doesn't recognise the "local" include of `bits.hpp` in thi file, so `bits.hpp` gets grouped with all othe rincludes from the library.  In `.cpp` files, clang-format understands the file extension and therefore *does* recognise local headers, hence this separation between `bits.hpp` and other headers.

dtebbs:
All comments above should have been addressed now (commenting here rather than on all individual threads).

AntoineRondelet:
LGTM

PR #208, opened 2020-05-05
Tooling improvement

AntoineRondelet:
Adding several tools to the build in order to generate the project documentation and coverage reports, enforce good practices more aggressively and have access to security analysis tools to inspect our binaries.

Tools added:
- doxygen
- cppcheck
- include-what-you-use
- clang-tidy
- sanitizers (memory, leaks, thread, etc)
- lcov

PR #209, opened 2020-05-06
libzeth tests structure

dtebbs:
- Test file layout to match libzeth code
- Added tests for several components (more required)
- Fixed some duplicated code, ambiguous interfaces (particularly around endianness) and inefficient implementations.


libzeth/core/field_element_utils.hpp
```diff
@@ -52,14 +52,15 @@ std::string bigint_to_hex(const libff::bigint<FieldT::num_limbs> &limbs);
 /// and we have `res_bigint = [Limb0, Limb1, Limb2, Limb3]`,
 /// where every Limb{i} is of type `mp_limb_t`,
 template<typename FieldT>
-libff::bigint<FieldT::num_limbs> bigint_from_bytes(
-    const uint8_t bytes[(FieldT::num_bits + 8 - 1) / 8]);
-
-/// Convert an hexadecimal string to a field element
-template<typename FieldT> FieldT field_element_from_hex(std::string field_str);
+libff::bigint<FieldT::num_limbs> bigint_from_hex(const std::string &hex);
```
AntoineRondelet, line 55: I am not sure the comments above are still applicable to the changes you have made to the encoding functions. If I get it correctly, what you will have if you pass: `input_bigint = [Limb0, Limb1, Limb2, Limb3]` to your encoding function  `bigint_to_hex` and the result you should get is:
```
res = [x7, ..., x0, x15, ..., x8, x23, ..., x16, x31, ..., x24]
            Limb3       Limb2           Limb1         Limb0
```

i.e. the byte endianness is switched compared to what's written in the comment -> I preserved the byte endianness in the previous encoding by doing right shifts to select the bytes https://github.com/clearmatics/zeth/blob/develop/libzeth/libsnark_helpers/debug_helpers.tcc#L53-L54

AntoineRondelet: This should be fine because you used both "reverse" functions in your encoding "from" and "to" hex, so you have inverse functions. However, it is worth reflecting these changes to the comments otherwise this is confusing.

libzeth/test/CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
...
+# Most tests are considered FAST
+file(
+  GLOB_RECURSE
+  TEST_SOURCE_FILES
+  core/**_test.cpp
+  snarks/**_test.cpp
+  mpc/**_test.cpp
+  circuits/**_test.cpp
+  serialization/**_test.cpp
+  )
```
AntoineRondelet, line 70: Remove indent

libzeth/test/CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
...
+  add_executable(${TEST_NAME} EXCLUDE_FROM_ALL ${test_src})
+  target_link_libraries(
+    ${TEST_NAME}
+
+    zeth
+    gtest_main
+    ${Boost_SYSTEM_LIBRARY}
+    ${Boost_FILESYSTEM_LIBRARY}
+    protobuf::libprotobuf
+    )
```
AntoineRondelet, line 28: Remove indent

libzeth/test/circuits/sha256_test.cpp
```diff
@@ -63,36 +62,40 @@ TEST(TestSHA256, TestHash)
...
-         1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1,
-         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1},
-        ZERO);
+    libsnark::pb_variable_array<FieldT> left =
+        libzeth::variable_array_from_bit_vector(
+            {0, 0, 0, 0, 1, 1, 1, 1, // 0, 0, 0, 0, 1, 1, 1, 1,
```
AntoineRondelet, line 67: Let's either comment the hexadecimal representation of the byte `// 0x0F` or `0x00` after each line, or just remove this `// 0, 0, 0, 0, 1, 1, 1, 1,` comment

libzeth/test/circuits/sha256_test.cpp
```diff
@@ -63,36 +62,40 @@ TEST(TestSHA256, TestHash)
...
-         1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1,
-         1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1},
-        ZERO);
+    libsnark::pb_variable_array<FieldT> right =
+        libzeth::variable_array_from_bit_vector(
+            {0, 1, 0, 0, 0, 0, 1, 1, //  (0, 1 is the right prefix here)
```
AntoineRondelet, line 85: Same here (see comment above)

libzeth/test/circuits/sha256_test.cpp
```diff
@@ -2,23 +2,22 @@
...
+#include "libzeth/circuits/circuit_types.hpp"
+#include "libzeth/circuits/circuit_utils.hpp"
+#include "libzeth/circuits/sha256/sha256_ethereum.hpp"
 
 #include <gtest/gtest.h>
 #include <libsnark/common/data_structures/merkle_tree.hpp>
 
 using namespace libsnark;
-using namespace libzeth;
+// using namespace libzeth;
```
AntoineRondelet, line 14: Let's remove this comment if we do not want to use the namespace here.
Moreover, let's make this consistent across test files as for now some are using the namespace and some aren't

AntoineRondelet: I am happy to remove `using libzeth`. I tend to prefer this approach anyway (I really do not like overusing `using <namespace>` since you lose track of where the function comes from). That being said I do not have a strong preference for this specific scenario: both approaches are "equally good to me" for small test files like these ones. So let's pick one and stick to it consistently

libzeth/test/core/utils_test.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include "libzeth/core/utils.hpp"
+
+#include <gtest/gtest.h>
+
+namespace
+{
+
+static const size_t DUMMY_BUFFER_SIZE = 16;
+
+static const uint8_t dummy_bytes[DUMMY_BUFFER_SIZE] = {
```
AntoineRondelet, line 14: Remark: We need to be consistent with styling. Here there are 2 static consts but 2 naming styles (snake_case and UPPER_CASE).
We will deal with these in another PR once https://github.com/clearmatics/zeth/pull/208 is merged since I have started to configure all these things in the clang-tidy config file.
Spoiler: I have set the naming convention for static const to be UPPER_CASE

libzeth/core/utils.cpp
```diff
@@ -83,28 +112,21 @@ std::string bytes_to_hex(const std::string &s)
...
 
-void erase_substring(std::string &string, const std::string &substring)
-{
-    size_t position = string.find(substring);
+    const uint8_t *const src_bytes_end = (const uint8_t *)bytes;
+    const uint8_t *src_bytes = src_bytes_end + num_bytes;
+    do {
+        --src_bytes;
+        const uint8_t byte = *src_bytes;
+        out.push_back(nibble_hex(byte >> 4));
```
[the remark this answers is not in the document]
dtebbs: The reason I don't mask is that it is redundant because the shift  `>> 4` is required.

libzeth/core/utils.cpp
```diff
@@ -83,28 +112,21 @@ std::string bytes_to_hex(const std::string &s)
     return out;
 }
 
-int hex_to_bytes(char *source_str, uint8_t *dest_buffer)
+std::string bytes_to_hex_reversed(const void *bytes, size_t num_bytes)
```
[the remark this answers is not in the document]
dtebbs: It's pretty common to use`void *` in these situations when reading/writing bytes from a range of underlying data formats.  It avoids having to cast from various pointer types (the underlying data type is quite unlikely to actually be`uint8_t`).

libzeth/core/utils.cpp
```diff
@@ -83,28 +112,21 @@ std::string bytes_to_hex(const std::string &s)
...
-{
-    size_t position = string.find(substring);
+    const uint8_t *const src_bytes_end = (const uint8_t *)bytes;
+    const uint8_t *src_bytes = src_bytes_end + num_bytes;
+    do {
+        --src_bytes;
+        const uint8_t byte = *src_bytes;
+        out.push_back(nibble_hex(byte >> 4));
+        out.push_back(nibble_hex(byte & 0x0f));
+    } while (src_bytes > src_bytes_end);
```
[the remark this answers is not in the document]
dtebbs: You mean the do-while?  It can be a little less awkward to do decrementing loops like this. We don't have to setup the pointer as `dest_bytes + bytes - 1` or use the `>=` condition which is less safe with indices and pointers. (admittedly, with pointers it would only be an issue if `src_bytes_end == 0` which should never happen on any regular platform, but in general this form seemed more sensible when decrementing).

(Let me know if you are referring to something else here)

libzeth/CMakeLists.txt
```diff
@@ -70,70 +71,7 @@ target_link_libraries(
...
-  endif()
-
-  # Extra dependencies for specific tests
-  target_link_libraries(
-    test_prover
-    ${Boost_SYSTEM_LIBRARY}
-    ${Boost_FILESYSTEM_LIBRARY}
-  )
-
+  add_subdirectory(test)
```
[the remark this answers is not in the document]
dtebbs: Moved

libzeth/core/utils.cpp
```diff
@@ -40,41 +46,64 @@ static char nibble_hex(const uint8_t nibble)
     return '0' + nibble;
 }
 
-std::string hex_to_bytes(const std::string &s)
+// Return a pointer to the beginning of the actual hex characters (removing any
+// `0x` prefix), and ensure that the length is as expected.
+static const char *find_hex_start(const std::string &hex, const size_t bytes)
```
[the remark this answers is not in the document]
dtebbs: I prefer to avoid `sanitize` here since that implies we are acting on the string itself and that we are doing some kind of thorough check (e.g. that all characters are valid hex).  I've renamed to `find_hex_string_of_length`.  Let me know if you prefer something else - I may have misunderstood the original objection.

libzeth/core/field_element_utils.hpp
dtebbs: I hadn't intended to change the encoding here.  I'll check this and fix it.

libzeth/core/utils.cpp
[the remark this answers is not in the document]
AntoineRondelet: Ah right, I see. Your `nibble_hex` function expects to receive something shifted (as illustrated by the assert here: https://github.com/clearmatics/zeth/pull/209/files#diff-2e27b8cb551e7bf14afeb41a4ebb67c7L35)
Makes sense

libzeth/core/utils.cpp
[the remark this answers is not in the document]
AntoineRondelet: > You mean the do-while?

Yes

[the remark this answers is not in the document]
AntoineRondelet: As pointed out by @dtebbs `i` is `size_t` so the condition above is always true for unsigned ints. We'd need to switch `i` to be signed for the loop to end

libzeth/core/utils.cpp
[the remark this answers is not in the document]
AntoineRondelet: Was suggesting to cast to `uint8_t*` before calling the function, but as discussed let's keep it that way

libzeth/core/field_element_utils.hpp
dtebbs: (As discussed, I realised the new code was equivalent to the old code afterall.  Updated the comments a little bit.)

libzeth/test/CMakeLists.txt
dtebbs: removed

libzeth/test/circuits/sha256_test.cpp
dtebbs: removed the comment and formatted more logically (16 bits per line)

libzeth/test/circuits/sha256_test.cpp
dtebbs: reformatted

libzeth/test/circuits/sha256_test.cpp
dtebbs: Removed the comment.

libzeth/test/core/utils_test.cpp
dtebbs: I think there is scope for a difference between genuine "constants" (traditionally represented as `#define`s with all caps, which may completely disappear at compile time - e.g. the `const size_t` here) and structured data consisting of constant values.
As discussed, let's see what the linter tools flag / allow and make a decision.

libzeth/core/field_element_utils.hpp
AntoineRondelet: Thanks

dtebbs:
@AntoineRondelet thanks for the comments.  I believe everything is addressed, but let me know if there are any remaining points.

libzeth/test/core/utils_test.cpp
AntoineRondelet: Yes, the tools provides an extensive set of possible configurations (here's a subset of such checks for identifier naming: https://clang.llvm.org/extra/clang-tidy/checks/readability-identifier-naming.html).
Let's refine this in a subsequent PR, and iterate on the coding conventions

Issue #191, opened 2020-04-08
Replacing pointers to constant references.


rrtoledo:
See. https://github.com/clearmatics/zeth/pull/157#discussion_r402451322

PR #210, opened 2020-05-06
libzeth signature consistency

dtebbs:
Use const refs (where applicable) for passing complex classes.
(`pb_variable` is a special case where there is no advantage to passing by const ref, but changed to const ref everywhere for consistency).


AntoineRondelet:
Is this the PR associated to https://github.com/clearmatics/zeth/issues/191 @dtebbs ?

libzeth/CMakeLists.txt
```diff
@@ -72,4 +72,6 @@ target_link_libraries(
 add_dependencies(zeth libsodium)
 
 # Tests
-add_subdirectory(test)
+if ("${IS_ZETH_PARENT}")
+  add_subdirectory(test)
```
AntoineRondelet, line 76: tests

dtebbs:
> Is this the PR associated to #191 @dtebbs ?

Yes, I think it does in a way.  That issues refers to pointers vs const refs, although I don't think we have many raw pointers any more, so I guess that can be closed when this is merged.

AntoineRondelet:
> Yes, I think it does in a way. That issues refers to pointers vs const refs, although I don't think we have many raw pointers any more, so I guess that can be closed when this is merged.

We still pass some `shared_ptr` by value to some functions. I would assume that this induces some runtime overhead. I'd need to look into that, so I will keep the associated issue (#191) open for the time being.

dtebbs:
> We still pass some shared_ptr by value to some functions. I would assume that this induces some runtime overhead. I'd need to look into that, so I will keep the associated issue (#191) open for the time being.

Ah, I see.  I thought it meant raw pointers rather than smart pointer.  Yes, there are some `shared_pointers` still - I assumed they were used for memory management so wasnt' sure if they can just be replaced.  They have a tiny runtime overhead but it will be insignificant compared to the proof gen, so likely justifiable given the memory safety.  Let's check in a separate PR, and decide on the best way to manage that memory.

AntoineRondelet:
> > We still pass some shared_ptr by value to some functions. I would assume that this induces some runtime overhead. I'd need to look into that, so I will keep the associated issue (#191) open for the time being.
> 
> Ah, I see. I thought it meant raw pointers rather than smart pointer. Yes, there are some `shared_pointers` still - I assumed they were used for memory management so wasnt' sure if they can just be replaced. They have a tiny runtime overhead but it will be insignificant compared to the proof gen, so likely justifiable given the memory safety. Let's check in a separate PR, and decide on the best way to manage that memory.

Yes definitely no compromise possible on memory safety. I am keen to see if there is a more efficient way to pass our gadgets around though.

PR #212, opened 2020-05-06
[WIP] Clang tidy fix

AntoineRondelet:

PR #213, opened 2020-05-07
Incremental clang-tidy fixes

dtebbs:
Related to #212.  Proposal for a subset of warnings to enable for now.  Many of the disabled warnings are worth addressing, but may be better done later.


prover_server/prover_server.cpp
```diff
@@ -406,7 +406,7 @@ int main(int argc, char **argv)
 
 #ifdef DEBUG
     // Run only if the flag is set
-    if (jr1cs_file != "") {
+    if (jr1cs_file.empty()) {
```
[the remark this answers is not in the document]
dtebbs: That's my mistake.  Thanks.

.clang-tidy
```diff
@@ -38,30 +36,47 @@ Checks: >
     readability-*,
     -modernize-avoid-c-arrays,
     -modernize-use-bool-literals,
-    -modernize-use-auto
+    -modernize-use-auto,
+    -modernize-use-trailing-return-type,
+    -readability-magic-numbers,
```
AntoineRondelet, line 41: Do we use "magic numbers" in many places? I like this check actually

.clang-tidy
```diff
@@ -38,30 +36,47 @@ Checks: >
...
+    -readability-implicit-bool-conversion,
+    -modernize-return-braced-init-list,
+    -readability-named-parameter,
```
AntoineRondelet, line 44: Do we use functions with unnamed arguments?

.clang-tidy
```diff
@@ -38,30 +36,47 @@ Checks: >
...
+    -modernize-pass-by-value,
+    -modernize-use-default-member-init,
+    -bugprone-exception-escape,
+    -misc-non-private-member-variables-in-classes,
+    -misc-unused-parameters,
+    -readability-redundant-member-init,
+    -modernize-deprecated-headers,
+    -readability-identifier-naming,
```
AntoineRondelet, line 52: I can see that you have disabled many checks here, which is fine (if we can't do otherwise for now). I think it would be nice to add a few lines about why each check has been disabled (as I started to do above) in order to make it crystal clear that we have a reason to remove the checks.

Eg. If we have a lot of magic numbers (in the tests for eg) we can add a line to say "-readability-magic-numbers is removed because we have a lot of magic numbers in the tests [...]". Likewise, I imagine that you removed `readability-implicit-bool-conversion` because of the boolean vectors we use in the hash implementation, right?

libzeth/circuits/blake2s/blake2s.hpp
```diff
@@ -25,66 +25,67 @@ template<typename FieldT> class BLAKE2s_256 : public libsnark::gadget<FieldT>
 {
 private:
     // Parameter block, size set to 32 bytes, fanout and depth set to serial
-    // mode. See: Section 2.8 https://blake2.net/blake2.pdf Table 2
-    std::array<std::array<bool, BLAKE2s_word_size>, 8> parameter_block = {
+    // mode. See: Section 2.8 https://blake2.net/blake2.pdf Table 2. Note, byte
+    // endianness has been swapped to avoid doing this repeatedly at runtime.
```
AntoineRondelet, line 29: What do you mean by "to avoid doing this repeatedly at runtime"?

dtebbs: Previously, these constants were endian-swappd before being used (see the .tcc file), and this swap was happening for every invocation of `generate_r1cs_witness` for every instance of the gadget.  Those constants did not seem to be used anywhere else, so the byte swapping seemed unnecessary.

> What do you mean by "to avoid doing this repeatedly at runtime"?

It refers to the byte swapping.  It's done in the static data, so it doesn't have to be done many times when running the proof gen.  (I put the comment to for anyone looking at the reference above).

libzeth/core/bits.hpp
```diff
@@ -15,70 +17,66 @@
...
+    static bits from_size_t(size_t addr);
 
-/// "Construct" `bits` types from boolean vectors
-bits384 bits384_from_vector(const std::vector<bool> &vect);
+    bool is_zero() const;
 
-/// "Construct" `bits` types from hexadecimal strings
-bits384 bits384_from_hex(const std::string &hex_str);
+    template<typename FieldT>
+    void fill_variable_array(
```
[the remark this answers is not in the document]
dtebbs: Added a comment

.clang-tidy
dtebbs: Added some comments about the disabled checks.

.clang-tidy
dtebbs: Yes, in the case of unused parameters we remove the name to stop the compiler complaining.

.clang-tidy
dtebbs: In general it's pretty good, but picks up a few bits of test data and some numbers in expressions (masking expressions, etc) which don't really warrant a whole separate constant.  It would be nice if it was a tiny bit less sensitive, so we may want to revisit in the future.

AntoineRondelet: I see, thanks

.clang-tidy
AntoineRondelet: Ah yes right...

libzeth/core/bits.hpp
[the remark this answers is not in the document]
AntoineRondelet: Thanks

.clang-tidy
```diff
@@ -27,7 +25,43 @@
...
+#   changes. May wish to review in the future.
+#
+# -readability-magic-numbers
+#   Disabled for now as it picks up some constants which don't necessarily need
+#   their own variables (for some trivial multiplication and masking) and some
+#   test data. May wish to review in the future.
+#
+# -readability-implicit-bool-conversion
+#   We use implicit bool conversion for specifying many constants in terms of
+#   `0` and `1`. (TODO: it may be a good idea to enable this to fix implicit
```
AntoineRondelet, line 47: We can enable this check bu we will need to add a lot of `NOLINT` comments to disable the checks when we use pb_variable_arrays of booleans (like in blake/sha). This is what I did on https://github.com/clearmatics/zeth/pull/212

libzeth/circuits/blake2s/blake2s.hpp
AntoineRondelet: Ah yes, I see. The parameter block structure was taken from Table 2. https://blake2.net/blake2.pdf and since blake processed byte strings as little endian the endianness needed to be switched.
Yes that's fine to switch the endianness directly in the parameter block definition. This is what is done in the RFC (see section 2.5 https://tools.ietf.org/html/rfc7693#page-5)

PR #214, opened 2020-05-07
libzeth: fix segfault in filesystem_util

dtebbs:
fix segfault under some circumstances
rearranged code slightly so potentially null var is only handled in confined set of lines.

### DIFF
--- a/libzeth/serialization/filesystem_util.cpp
+++ b/libzeth/serialization/filesystem_util.cpp
@@ -10,33 +10,33 @@ namespace libzeth
 boost::filesystem::path get_path_to_setup_directory()
 {
     const char *path = std::getenv("ZETH_TRUSTED_SETUP_DIR");
-    if (nullptr == path) {
-        // Fallback destination if the ZETH_TRUSTED_SETUP_DIR env var is not set
-        // We assume below that `std::getenv("HOME")` does not return `nullptr`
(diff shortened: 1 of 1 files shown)


libzeth/serialization/filesystem_util.cpp
```diff
@@ -10,33 +10,33 @@ namespace libzeth
...
-    if (nullptr == path) {
-        // Fallback destination if the ZETH_TRUSTED_SETUP_DIR env var is not set
-        // We assume below that `std::getenv("HOME")` does not return `nullptr`
-        boost::filesystem::path home_path =
-            boost::filesystem::path(std::getenv("HOME"));
-        boost::filesystem::path zeth_setup("zeth_setup");
-        boost::filesystem::path default_path = home_path / zeth_setup;
-
-        return default_path;
+    if (nullptr != path) {
```
AntoineRondelet, line 13: Let's invert the 2 `path != nullptr`

AntoineRondelet:
Good catch @dtebbs, a return was indeed missing in one of the block. Thanks!

libzeth/serialization/filesystem_util.cpp
dtebbs: updated

AntoineRondelet: Thanks

PR #218, opened 2020-05-13
libzeth: fix json output of pghr13 proofs

dtebbs:
Fix pghr13 `proof_write_json` funnction, which was using the default stream operator (which produces binary by default).

### DIFF
--- a/libzeth/snarks/pghr13/pghr13_snark.tcc
+++ b/libzeth/snarks/pghr13/pghr13_snark.tcc
@@ -117,7 +117,17 @@ template<typename ppT>
 std::ostream &pghr13_snark<ppT>::proof_write_json(
     const typename pghr13_snark<ppT>::ProofT &proof, std::ostream &os)
 {
-    return os << proof;
+    os << "{\n";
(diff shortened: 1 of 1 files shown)


libzeth/snarks/pghr13/pghr13_snark.tcc
```diff
@@ -117,7 +117,17 @@ template<typename ppT>
 std::ostream &pghr13_snark<ppT>::proof_write_json(
     const typename pghr13_snark<ppT>::ProofT &proof, std::ostream &os)
 {
-    return os << proof;
+    os << "{\n";
+    os << " \"a\" :[" << point_g1_affine_to_json<ppT>(proof.g_A.g) << "],\n";
```
AntoineRondelet, line 121: I don't think you need the surrounding `[` and `]` here as these are already added in the `point_g1_affine_to_json` functions which write the group elements as an array of hex string https://github.com/clearmatics/zeth/blob/master/libzeth/core/group_element_utils.tcc#L18

dtebbs: Thanks.  I restored it from the old code and forgot to fix this.  Should be updated now.

PR #219, opened 2020-05-13
Added empty constructor for extended_proof class

AntoineRondelet:
In certain context, having an empty constructor for our classes is required (if we have `std::array` of objects of the given class). This PR adds the empty constructor for the `extended_proof` class.


AntoineRondelet:
Closing this PR as we use an array of extended_proof pointers in zecale (see: https://github.com/clearmatics/zecale/pull/15), which bypasses the need for an empty constructor here.

PR #225, opened 2020-06-18
Changes to support bls12-377 and bw6-761

dtebbs:
- Remove command line macros for curve and snark
- Library code is fully parameterized (does NOT include zeth_config.h or include any notion of "defaults")
- Application code uses configured snark and curve types form the zeth_config.h
- type variables use `ppT`, types use `pp`


CMakeLists.txt
```diff
@@ -39,38 +39,42 @@ elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
...
+  ZETH_CURVE
+  "ALT_BN128"
+  CACHE
+  STRING
+  "Default curve: one of BLS12_377, BW6_761, ALT_BN128, BN128, EDWARDS, MNT4, MNT6"
 )
 
-# Flags and compilation options to chose the type of zksnark
+# Configure a header file to pass some of the CMake settings
+# Option selecting the type of zksnark
```
AntoineRondelet, line 56: `Option selecting the type of zksnark` -> `Option selecting the zk-SNARK scheme`

libzeth/tests/circuits/commitments_test.cpp
```diff
@@ -60,7 +62,7 @@ TEST(TestCOMMs, TestCOMMGadget)
     libsnark::pb_variable<FieldT> result;
     result.allocate(pb, "result");
 
-    COMM_cm_gadget<FieldT, HashT> comm_cm_gadget(pb, a_pk, rho, r, v, result);
+    COMM_cm_gadget<FieldT, Hash> comm_cm_gadget(pb, a_pk, rho, r, v, result);
```
AntoineRondelet, line 65: Should we switch `FieldT` to `Field` then?

libzeth/tests/circuits/note_test.cpp
```diff
@@ -10,19 +10,17 @@
...
-using ppT = libzeth::ppT;
-
-// Should be alt_bn128 in the CMakeLists.txt
-using FieldT = libff::Fr<ppT>;
-
 // We use our hash functions to do the tests
-using HashT = BLAKE2s_256<FieldT>;
-using HashTreeT = MiMC_mp_gadget<FieldT>;
+using pp = defaults::pp;
+using FieldT = defaults::FieldT;
```
AntoineRondelet, line 21: Same question as above, re: `FieldT`->`Field`

dtebbs: Yes, technically we should.  I was hesitant because of the number of lines it might touch, but since you raised it I'll include this change.

dtebbs: I've switched this everywhere.  I think we should now be consistent with the type var vs type name convention.

CMakeLists.txt
dtebbs: changed

libzeth/tests/circuits/commitments_test.cpp
dtebbs: Switched to `Field` everywhere.

CMakeLists.txt
AntoineRondelet: Thanks!

libzeth/tests/circuits/note_test.cpp
AntoineRondelet: > Yes, technically we should. I was hesitant because of the number of lines it might touch, but since you raised it I'll include this change.

Thanks @dtebbs 
Yes, I think it's important to be consistent here

AntoineRondelet:
Looking good to me, great to see the library code fully parameterized @dtebbs!

Issue #228, opened 2020-07-20
MerkleTreeMiMC7_test.sol:24:25: Error: Array is too large to be encoded. public returns (bytes32[MAX_NUM_NODES] memory)


yanyanho:
when I ran test_merkle_tree_contract.py, I came across this problem.

AntoineRondelet:
Thanks for the issue @yanyanho, that's a good catch!
We used the protocol parameters for the tests - i.e. a Merkle tree of depth 32 - which is way to big to be used with the test function that returns the encoded tree.
This is fixed in https://github.com/clearmatics/zeth/pull/229, where I used some testing parameters to process a Merkle tree of depth 8 in the `test_merkle_tree_contract.py` test. I all works well on my end, please let me know if you have any further troubles.

yanyanho:
Thanks very much, I hava a further trouble.  A Merkle tree of depth 8 can only support 2^8 transfers,  can you give the demo dynamic creating merkel trees.     A Merkle tree of depth 32 takes too long time (20s , mac docker test ) to generating proof, I want to decrease the tree depth but I need to dynamic creating merkel trees.    What's more , Can you give some suggestion to short the time of generating proof.

dtebbs:
Hi @yanyanho.

The Merkle tree size is controlled by the `ZETH_MERKLE_TREE_DEPTH` constants in constants.py and zeth_constants.hpp, and `DEPTH` in BaseMerkleTree.sol.  These must all be set to the same value BEFORE the prover is built and the contract deployed.  (Unfortunately we do not yet have a nicer way of configuring this).  This affects the verification and proving keys for the ZK proofs, and so at present we do not have a practical way to dynamically change the depth of the tree.

Regarding proving times, it's worth checking the fundamentals (ensure you are running an optimized multi-threaded build and that the docker container has enough resources, or try a native mac build etc) to get the max performance, but a substantial improvement would require optimization of the arithmetic circuit, and even this is only possible up to a point.

You may be interested in Chapter 4 of the protocol spec https://github.com/clearmatics/zeth-specifications/blob/master/zeth-protocol-specification.pdf which discusses performance, snark-friendly primitives and some of the trade-offs in this proof-of-concept.

AntoineRondelet:
@yanyanho this ticket has been addressed in #229 which is now merged in `develop`, and which will then be introduced in the next release.
[...]

PR #229, opened 2020-07-21
Fix python tests

AntoineRondelet:
- The command `make test_contracts` in the `client` was failing because a function - tested - was previously removed, and the test never updated. Fixed that.

- Added the `make test` command of the `client` to the ci script

- Renamed functions of the ci script consistently

- Fixed `test_merkle_tree_contract.py` which was not working properly due to the merkle tree being too deep: Used a smaller Merkle tree for the test. Note: I needed to duplicate code in order to "repair" this test. This is obviously pretty bad, but seems to be a consequence of the way the `BaseMerkleTree.sol` contract has been implemented. In fact, the depth of the tree (i.e. `DEPTH`) is hardcoded in the contract, which makes the contract very hard to re-use across contexts. We already had this discussion @dtebbs - I wouldn't be against removing the `require` you added in the constructor - but let's discuss this further.


AntoineRondelet:
For some reason, I can't get the `make test` command to work on the ci server. I temporarily commented the `test-pyclient` job in the github action workflow, as I can't spend more time on that atm. Will need to get back to this later.

The error I get is copied/pasted below:
```bash
(env) bash-5.0# make test
python -m unittest
EEEE.................
======================================================================
ERROR: test_mix_parameters (test.test_contracts.TestContracts)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/zeth2/client/test/test_contracts.py", line 35, in test_mix_parameters
    receiver_enc_keypair = generate_encryption_keypair()
  File "/home/zeth2/client/zeth/encryption.py", line 164, in generate_encryption_keypair
    sk = generate_encryption_secret_key()
  File "/home/zeth2/client/zeth/encryption.py", line 107, in generate_encryption_secret_key
    return EncryptionSecretKey(X25519PrivateKey.generate())  # type: ignore
  File "/home/zeth2/client/env/lib/python3.8/site-packages/cryptography/hazmat/primitives/asymmetric/x25519.py", line 42, in generate
    raise UnsupportedAlgorithm(
cryptography.exceptions.UnsupportedAlgorithm: X25519 is not supported by this version of OpenSSL.

======================================================================
ERROR: test_decryption_invalid_key (test.test_encryption.TestEncryption)
Tests that ONLY the owner of the receiver key can decrypt the ciphertext.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/zeth2/client/test/test_encryption.py", line 36, in test_decryption_invalid_key
    sk_1 = encryption.decode_encryption_secret_key(_TEST_SECRET_KEY_1_BYTES)
  File "/home/zeth2/client/zeth/encryption.py", line 117, in decode_encryption_secret_key
    X25519PrivateKey.from_private_bytes(sk_bytes))
  File "/home/zeth2/client/env/lib/python3.8/site-packages/cryptography/hazmat/primitives/asymmetric/x25519.py", line 53, in from_private_bytes
    raise UnsupportedAlgorithm(
cryptography.exceptions.UnsupportedAlgorithm: X25519 is not supported by this version of OpenSSL.

======================================================================
ERROR: test_encrypt_decrypt (test.test_encryption.TestEncryption)
Tests the correct encrypt-decrypt flow: decrypt(encrypt(m)) == m
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/zeth2/client/test/test_encryption.py", line 26, in test_encrypt_decrypt
    sk = encryption.decode_encryption_secret_key(_TEST_SECRET_KEY_BYTES)
  File "/home/zeth2/client/zeth/encryption.py", line 117, in decode_encryption_secret_key
    X25519PrivateKey.from_private_bytes(sk_bytes))
  File "/home/zeth2/client/env/lib/python3.8/site-packages/cryptography/hazmat/primitives/asymmetric/x25519.py", line 53, in from_private_bytes
    raise UnsupportedAlgorithm(
cryptography.exceptions.UnsupportedAlgorithm: X25519 is not supported by this version of OpenSSL.

======================================================================
ERROR: test_private_key_generation (test.test_encryption.TestEncryption)
Sample some random keys and ensure that they comply with the
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/zeth2/client/test/test_encryption.py", line 49, in test_private_key_generation
    sk = encryption.generate_encryption_secret_key()
  File "/home/zeth2/client/zeth/encryption.py", line 107, in generate_encryption_secret_key
    return EncryptionSecretKey(X25519PrivateKey.generate())  # type: ignore
  File "/home/zeth2/client/env/lib/python3.8/site-packages/cryptography/hazmat/primitives/asymmetric/x25519.py", line 42, in generate
    raise UnsupportedAlgorithm(
cryptography.exceptions.UnsupportedAlgorithm: X25519 is not supported by this version of OpenSSL.

----------------------------------------------------------------------
Ran 21 tests in 1.559s

FAILED (errors=4)
make: *** [Makefile:40: test] Error 1
```

Not sure why `X25519` isn't supported by the version of openssl installed on Alpine. Need to spend more time on that when I can. 
Will keep this PR open for the time being and close it once this issue is resolved.

AntoineRondelet:
I'll open a separate ticket for the openssl issue, and tackle it later on.
This PR fixes #228

scripts/ci
```diff
@@ -21,19 +21,20 @@ function format_check() {
     scripts/check_copyright
 }
 
-function contracts_check() {
+function check_contracts() {
```
dtebbs, line 24: Swapping the order here means we have some functions called `<module>_check` and some called `check_<module>`.  Let's either rename all, or leave the original ordering.

scripts/ci
```diff
@@ -21,19 +21,20 @@ function format_check() {
...
     pushd zeth_contracts
     npm config set python python2.7
     npm install
     npm run check
     popd
 }
 
-function pyclient_check() {
+function check_pyclient() {
```
dtebbs, line 32: While this is being changed, let's change `pyclient` to `client`.

zeth_contracts/contracts/BaseMerkleTree_test.sol
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) 2015-2020 Clearmatics Technologies Ltd
```
dtebbs, line 1: If I've understood correctly, this duplicates all of the merkle tree contract code which is a bit concerning.
I think the original problem is that we try to return a set of nodes that is too large.
Could this not be fixed by having `MerkleTreeMiMC7_test::testAddLeaves` return just the new root and compare that to the version on the host?  I think this would still give us a high level of confidence that the contract was doing the right thing.

If we want to check specific nodes on the contract, perhaps `MerkleTreeMiMC7_test` could expose another method to return single nodes, or ranges of nodes, but this is probably not necessary if we check that the root is consistent.

scripts/ci
AntoineRondelet: Good one, I missed the format_check.
I'll rename it (I want these function names to be consistent with the check names in the ci script which wasn't the case befofre)

zeth_contracts/contracts/BaseMerkleTree_test.sol
AntoineRondelet: Yes I've switched to that. Don't know why the full set of nodes was returned before TBH...
The tests only require the value of `nodes[0]` (the root) to work.
It seems that we were trying to return the full set of tree nodes just to build an error message https://github.com/clearmatics/zeth/blob/develop/client/test_commands/test_merkle_tree_contract.py#L44

scripts/ci
AntoineRondelet: Indeed. Switched to that

PR #233, opened 2020-07-27
Consistent naming of types and type-params

dtebbs:


libzeth/snarks/pghr13/pghr13_snark.tcc
```diff
@@ -34,7 +34,7 @@ typename pghr13_snark<ppT>::ProofT pghr13_snark<ppT>::generate_proof(
 
     // Generate proof from public input, auxiliary input (private/secret data),
     // and proving key
-    ProofT proof = libsnark::r1cs_ppzksnark_prover(
+    proof proof = libsnark::r1cs_ppzksnark_prover(
```
AntoineRondelet, line 37: This can be confusing, we need to be careful here. One `proof` is the "type alias" defined with a "using" directive, the other `proof` is a variable name.
So the proposed convention is (if I follow correctly):
1. Type params: CamelCase + finishing with `T`
2. Type aliases (`using`): snake_case with no suffix

Happy with that, but we need to be a bit careful here because there are no "special pattern" used to differentiate the "type aliases", and since these are not allocated keywords, we will certainly have collisions in names throughout the code (exactly as the line above).

dtebbs: >    Type params: CamelCase + finishing with T
>    Type aliases (using): snake_case with no suffix

Concrete types, including aliases, are snake case (i.e. consistent with the current class naming where this exact same thing could technically occur).  Type parameters have a trailing `T` and I've tried to use lower case in most cases to match the current `ppT`, `snarkT`, `snarkT` (which seems more consistent with our concrete type naming than CamelCase), etc, although I've kept names like `FieldT` (which matches libsnark etc). I don't think it's critical, but we could change everything to a single convention.  But any convention will conflict with the `ppT` and `FieldT` used in libsnark and libff.

AntoineRondelet: > I don't think it's critical, but we could change everything to a single convention.

Not a fan of that, this will contrast too much with existing code in the libraries we use.
That sounds good. Could you add a bullet to the CODING_STANDARDS (https://github.com/clearmatics/zeth/blob/develop/CODING_STANDARDS.md#naming-and-ordering-conventions)? to account for the concrete types, i.e. change the entry: `Variables, functions and classes use snake_case_names.` to broaden it to "concrete types", i.e. classes and type aliases

dtebbs: > >    I don't think it's critical, but we could change everything to a single convention.
>
> Not a fan of that, this will contrast too much with existing code in the libraries we use.

Exactly.

I've updated the coding standards file (and also removed the `proof` temporary in the code above, since we didn't actually need an intermediate variable there).

AntoineRondelet: > I've updated the coding standards 

Thanks

> also removed the proof temporary in the code above, since we didn't actually need an intermediate variable there

Indeed

PR #239, opened 2020-08-12
WIP: preliminary support for local keys

dtebbs:
- [X] use `eth-private-key` if present to locally sign the tx (as well as `eth-address` as before)
   - [X] test commands/gen_eth_address.py generates `eth-private-key` and `eth-address` in the current dir
   - [X] test_commands/fund_eth_address.py funds the address, (assumes an the address is unlocked, funds from `web3.eth.accounts[0]` unless told otherwise) assuming gnache / geth with hosted accounts.
   - [X] test scripts (e.g. `scripts/test_zeth_cli` call `setup_user` in `scripts/test_zeth_client_common.sh` to extract a (funded) hosted address from the RPC host. Create two setup functions `setup_user_local_key` (userd for alice, bob, charlie in the test script) and `setup_user_hosted_key` (used for deployer in the test scripts)
   - [ ] Make deploy command also support `eth-private-key`
   - [ ] Create a script `fund_eth_address_autonity_test_net.py` or add flag to `fund_eth_address.py`, to fund `eth-address` from the test net faucet
   - [ ] [LOW PRIORITY] Make get_balances in the test scripts (`scripts/test_zeth_client_common.sh`) use the correct addresses

PR #241, opened 2020-08-21
Client changes to suppot geth-style nodes, such as autonity (depends on #239)

dtebbs:
Uses infraastructure in #239 to add a `autonity-helloworld` network configuration.
Support all client operations on `autonity-helloworld`, ensuring that `scripts/test_zeth_cli` and `scripts/test_zeth_cli_token` succeed.


AntoineRondelet:
LGTM

Issue #234, opened 2020-07-28
Issues with openssl in Python tests on Alpine


AntoineRondelet:
See: https://github.com/clearmatics/zeth/pull/229#issuecomment-661838284 for more background on that.

As explained in the [cryptography package documentation](https://readthedocs.org/projects/cryptography/downloads/pdf/stable/):
> Some Linux distributions (most notably Alpine) are not manylinux compatible so we cannot distribute wheels for them

As such, it is necessary to build a static wheels for the cryptography package.
I have done this locally and this fixes the issue, however, building the wheels is quite slow and will slow down the CI quite a lot.

Running:
```bash
#!/bin/bash
set -e
OPENSSL_VERSION="1.1.1g"

CWD=$(pwd)
virtualenv env
. env/bin/activate
pip install -U setuptools
pip install -U wheel pip
curl -O https://www.openssl.org/source/openssl-${OPENSSL_VERSION}.tar.gz
tar xvf openssl-${OPENSSL_VERSION}.tar.gz
cd openssl-${OPENSSL_VERSION}
./config no-shared no-ssl2 no-ssl3 -fPIC --prefix=${CWD}/openssl
make && make install
cd ..
CFLAGS="-I${CWD}/openssl/include" LDFLAGS="-L${CWD}/openssl/lib" pip wheel --wheel-dir wheels --no-binary :all: cryptography
```

Creates the wheels for the cryptography package, using the right version of openssl, under the `wheels` folder, which can then be installed with pip: `pip install --no-index --find-links=wheels cryptography` before running `make setup && make test` in the `test_pyclient` function in the ci script.

Allocate more thoughts into the CI and how to improve testing there. The main reason Alpine is used as a base image is to keep the zeth image as small as possible (Alpine being very lightweight). Moreover, we may want to ship an image for the prover only (no `client` related code, only the `prover_server` binary running and listening on a port mapped to the host), and have another image containing everything necessary to run the CI script (cpp + python code and tests etc).

This ticket is related to https://github.com/clearmatics/zeth/issues/150

AntoineRondelet:
Note that, a `constraints.txt` file needs to be used when building the wheel with `pip wheel` to make sure that the right version is built. If this is not specified, chances are that the docker image built will - each time a new release of `cryptography` is avail - fails mysteriously because the `setup.py` file requires a given version, while `pip wheel` command installs the latest (hence the `make setup` command will try to re-install the appropriate version and fail).
[...]

Issue #243, opened 2020-08-24
Look into supporting static builds


AntoineRondelet:
For now, the `prover_server` is dynamically built. It may be valuable to add options to the CMakeLists.txt files to allow for static build.
This obviously has cons: bigger binaries, less portable etc. That being said, that could be helpful to distribute the binaries packaged in a slim docker image.

AntoineRondelet:
The `prover_server` can now be statically built (added in #245), and the libzeth library is static by default. Closing this issue.

PR #245, opened 2020-08-24
Docker cleanup

AntoineRondelet:
Package all parts of the code base into their own docker images. This PR introduces:
- `Dockerfile-dev` (formerly `Dockerfile-zeth`): That copies all the zeth repo into an pre-configured alpine container, and which can be used for development matters and for the CI
- `Dockerfile-prover`: Build the `prover_server` and moves the binary into a slim alpine container. The associated image is only useful for people willing to run the `prover_server` only
- `Dockerfile-client`: Moves the client code into an alpine container. The resulting image can be used by people willing to run the client code only. This docker file also builds a static wheels for the `cryptography` package as per https://github.com/clearmatics/zeth/issues/234
- `Dockerfile-mpc`: Provides a slim docker image for MPC contributors


AntoineRondelet:
Partially tackling https://github.com/clearmatics/zeth/issues/243

Dockerfile-client
```diff
@@ -0,0 +1,53 @@
...
+    libffi-dev \
+    python3-dev \
+    py3-virtualenv \
+    bash
+
+# Copy necessary files in the docker container
+COPY ./client /home/zeth
+
+ENV OPENSSL_VERSION="1.1.1g"
+ENV CWD="/home/zeth"
```
dtebbs, line 20: Should this be something like `BUILD_ROOT` or similar?  I had to pause and understand since `CWD` is so strongly tied to current working directory.

dtebbs (changes requested):
Looks good. One small naming thing.

Dockerfile-client
AntoineRondelet: See: https://github.com/clearmatics/zeth/pull/245/commits/bdf9b884f5d7c190635f18b8c63018b290e57225

dtebbs (approved):

Issue #242, opened 2020-08-21
Setup serialization isn't compliant with deserialization on `prover_server`


AntoineRondelet:
If the `prover_server` is used in "dev-mode" , the server starts a local/trusted setup to generate an SRS.
After generating the SRS, it is written in 3 files: `pk.raw`, `vk.raw` and `vk.json` (this json form is useful for the contract-side).

Likewise, the `prover_server` has a flag to load a SRS (which should be used when not in dev-mode to use a SRS generated during an MPC). This flag expects one file.

There is an inconsistency here since, the setup is written over 2 files (ignoring the vk.json) at start up time of the prover, but loading a keypair expects 1 file. I think that would make more sense to be able to use one of the files of the setup serialization with the flag `--keypair` which is currently not the case.

AntoineRondelet:
Closing this issue as the corresponding PR has been merged

PR #248, opened 2020-08-25
Prover server flags

dtebbs:
Fix up prover server flags and output, so that keypair files can be written / read and debug output controlled fully.
Related to #242

### DIFF
--- a/prover_server/prover_server.cpp
+++ b/prover_server/prover_server.cpp
@@ -41,50 +41,35 @@ using circuit_wrapper = libzeth::circuit_wrapper<
 namespace proto = google::protobuf;
 namespace po = boost::program_options;
 
-static void serialize_setup_to_file(
-    const typename snark::keypair &keypair,
-    boost::filesystem::path setup_path = "")
+static snark::keypair load_keypair(const boost::filesystem::path &keypair_file)
 {
-    if (setup_path.empty()) {
-        setup_path = libzeth::get_path_to_setup_directory();
-    }
+    std::ifstream in_s(
+        keypair_file.c_str(), std::ios_base::in | std::ios_base::binary);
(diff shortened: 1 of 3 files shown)


prover_server/prover_server.cpp
```diff
@@ -282,36 +278,31 @@ static void RunServer(
...
 {
     // Options
     po::options_description options("");
     options.add_options()(
-        "keypair,k", po::value<std::string>(), "file to load keypair from");
-#ifdef DEBUG
+        "keypair,k",
+        po::value<std::string>(),
+        "file to load keypair from. If it doesn't exist, a new keypair will be "
+        "generated and written to this file. (defaults to [");
```
AntoineRondelet, line 297: The default should be $HOME assuming we run on a POSIX compliant OS

prover_server/prover_server.cpp
```diff
@@ -41,50 +41,35 @@ using circuit_wrapper = libzeth::circuit_wrapper<
...
+        keypair_file.c_str(), std::ios_base::in | std::ios_base::binary);
+    in_s.exceptions(
+        std::ios_base::eofbit | std::ios_base::badbit | std::ios_base::failbit);
+    return snark::keypair_read_bytes(in_s);
+}
 
-    // Write the verification key in json format
-    {
-        std::ofstream vk_json_s(path_vk_json.c_str());
-        snark::verification_key_write_json(verification_key, vk_json_s);
```
AntoineRondelet, line 62: Don't you write the VK in json format anymore? We need it to initialize the contracts

prover_server/prover_server.cpp
dtebbs: Updated the help.  Thanks.

PR #254, opened 2020-08-26
Refactor cli structure

AntoineRondelet:
- Refactored the structure of the CLIs
- Renamed `zeth_debug` CLI to `zeth_helper`
- Re-arranged the client code:
    - `zeth` becomes `core` (the cli is called `zeth`, enhance having a `zeth` folder suggested it was the code of the cli while it actually was the "core"/"backend" of the client)
    - `test` becomes `tests`
    - `commands` becomes `cli`
    - `test_commands` is now split in:
           - `test_commands`: scripts used to carry out "integration" tests and deploy testing contracts/generate data for the tests
           - `helper`: CLI allowing to carry out non-zeth related things. This is not strictly used for tests, but rather to ease the use of `Zeth` on an Ethereum-like blockchain by exposing some useful functionalities

- `core`, `api`, `helper`, `cli` are bundled into a `zeth` package


AntoineRondelet:
Supersedes #252

AntoineRondelet:
For some reason, this branch contains commits from the source branch that do not disappear when I try to rebase

dtebbs:
Just a few comments up front:

> zeth becomes core (the cli is called zeth, enhance having a zeth folder suggested it was the code of the cli while it actually was the "core"/"backend" of the client)
It woudl be greate if you could put everything in a `zeth` subdirectory, like we do in zecale.  This makes the hierarchy deep (so we have `client/zeth/core`, `client/zeth/commands` etc, but seems to be standard practice for python and it makes it much easier to use the code as a dependency (we can `import zeth.core` instead of `import core`)
I hit this kind of thing in several projects that use zeth, and meant to add this structure to zeth.  If you are renaming modules than now might be a good time to do it.

> commands becomes zeth_cli
So this can become `zeth/cli`, etc...  (The scripts entry in setup.py can refer to zeth/cli/zeth of course)

> test becomes unit_tests
I think `test` is a reserved name, known to the unit test framework and the package manager (dependencies cannot import the test module, I think), so it might be better to keep this as it is.

PR #259, opened 2020-08-28
Renamed environment variables and removed notestore

AntoineRondelet:


setup_env.sh
```diff
@@ -4,11 +4,8 @@ export ZETH=`pwd`
 export ZETH_CONTRACTS_DIR=$ZETH/zeth_contracts/contracts
 export ZETH_DEBUG_DIR=$ZETH/debug
 
-mkdir -p $ZETH/notestore
-export ZETH_NOTESTORE=$ZETH/notestore
-
-mkdir -p $ZETH/trusted_setup
-export ZETH_TRUSTED_SETUP_DIR=$ZETH/trusted_setup
+mkdir -p $ZETH/zeth_setup
+export ZETH_SETUP_DIR=$ZETH/zeth_setup
```
dtebbs, line 8: Not strictly part of this change, and just a small point, but the following would avoid duplicating the value of `ZETH_SETUP_DIR`
```
export ZETH_SETUP_DIR=$ZETH/zeth_setup
mkdir -p $ZETH_SETUP_DIR
```

dtebbs (commented):
Looks good.  Just one small suggestion.

setup_env.sh
AntoineRondelet: Indeed, that's a good point, thanks.

PR #263, opened 2020-09-01
Zeth client show parameters

dtebbs:
`mix --show-parameters` flag and related refactor.
fix for small output inconsistency when using `mix --wait`

### DIFF
--- a/client/zeth/cli/zeth_mix.py
+++ b/client/zeth/cli/zeth_mix.py
@@ -26,6 +26,7 @@ from typing import List, Tuple, Optional
 @option("--eth-addr", help="Sender's eth address or address filename")
 @option("--eth-private-key", help="Sender's eth private key file")
 @option("--wait", is_flag=True, help="Wait for transaction to be mined")
+@option("--show-parameters", is_flag=True, help="Show the mixer parameters")
 @pass_context
 def mix(
         ctx: Context,
@@ -35,7 +36,8 @@ def mix(
         output_specs: List[str],
         eth_addr: Optional[str],
         eth_private_key: Optional[str],
-        wait: bool) -> None:
+        wait: bool,
+        show_parameters: bool) -> None:
     """
     Generic mix function
     """
@@ -71,21 +73,27 @@ def mix(
     eth_address = load_eth_address(eth_addr)
     eth_private_key_data = load_eth_private_key(eth_private_key)
 
-    # If instance uses an ERC20 token, tx_value can be 0 not default vin_pub.
-    tx_value: Optional[EtherValue] = None
-    if mixer_desc.token:
-        tx_value = EtherValue(0)
+    # If instance uses an ERC20 token, tx_value can be 0. Otherwise it should
+    # match vin_pub.
+    tx_value = EtherValue(0) if mixer_desc.token else vin_pub
 
-    tx_hash = zeth_client.joinsplit(
+    mix_params = zeth_client.create_mix_parameters(
         wallet.merkle_tree,
         zeth_address.ownership_keypair(),
         eth_address,
-        eth_private_key_data,
         inputs,
         outputs,
         vin_pub,
-        vout_pub,
-        tx_value)
+        vout_pub)
+
+    if show_parameters:
+        print(f"mix_params={mix_params.to_json()}")
+
+    tx_hash = zeth_client.mix(
+        mix_params=mix_params,
+        sender_eth_address=eth_address,
+        sender_eth_private_key=eth_private_key_data,
+        tx_value=tx_value)
 
     print(tx_hash)
     if wait:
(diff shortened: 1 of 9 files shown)


client/zeth/cli/zeth_mix.py
```diff
@@ -26,6 +26,7 @@
 @option("--eth-addr", help="Sender's eth address or address filename")
 @option("--eth-private-key", help="Sender's eth private key file")
 @option("--wait", is_flag=True, help="Wait for transaction to be mined")
+@option("--show-parameters", is_flag=True, help="Show parameters to the mixer")
```
AntoineRondelet, line 29: `Show parameters to the mixer` -> `Show parameters of the mix call`?

client/zeth/cli/zeth_mix.py
```diff
@@ -75,18 +77,27 @@ def mix(
     tx_value: Optional[EtherValue] = None
     if mixer_desc.token:
         tx_value = EtherValue(0)
+    else:
+        tx_value = tx_value or vin_pub
```
AntoineRondelet, line 81: What is the purpose of this line?

client/zeth/cli/zeth_mix.py
```diff
@@ -75,18 +77,27 @@ def mix(
...
-        eth_private_key_data,
         inputs,
         outputs,
         vin_pub,
-        vout_pub,
-        tx_value)
+        vout_pub)
+
+    if show_parameters:
+        print(f"mix_params={mix_params.to_json()}")
```
AntoineRondelet, line 93: Is this flag used for debug purposes?

client/zeth/core/mixer_client.py
```diff
@@ -515,16 +515,17 @@ def mix(
             self,
             mix_params: contracts.MixParameters,
             sender_eth_address: str,
-            pub_value: Optional[EtherValue] = None,
+            sender_eth_private_key: Optional[bytes],
```
AntoineRondelet, line 518: That's a general note, not related to this PR in particular.

As a general rule of thumb, passing secret keys/data around is not desirable (at all). That's something to be careful with (I comment here because I can see that we pass the private key to functions here)

client/zeth/cli/zeth_mix.py
dtebbs: reworded this

client/zeth/cli/zeth_mix.py
dtebbs: The purpose is to set the ether value of the transaction based on nthe vout value.
This block is now a bit awkward after a refactor.  I've restructured it.

client/zeth/core/mixer_client.py
dtebbs: Sure. We can iterate in a future pass.

client/zeth/cli/zeth_mix.py
AntoineRondelet: >  I've restructured it.

Thanks for that, that's clearer indeed

client/zeth/cli/zeth_mix.py
AntoineRondelet: You wrote:
> @option("--show-parameters", is_flag=True, help="Show the mixer parameters")

I haven't ran this code, but if I followed it correctly, this flag seems to be used to display the args of `zeth mix` - not the parameters of the `Mixer` contract. Did I miss something?

dtebbs: It should dump the (json representation of) the `MixParameters` object, which is the in-memory version of the parameters passed to the mixer contract.

client/zeth/cli/zeth_mix.py
dtebbs: The original idea was that for docs / demos it would be nice to be able to show the info sent to the contract (i.e. the data visible to the blockchain), but I am sure it will be useful for debugging purposes as well.

client/zeth/cli/zeth_mix.py
AntoineRondelet: I see, shall `Show the mixer parameters` be changed to `Show the mix call parameters` then?

client/zeth/cli/zeth_mix.py
AntoineRondelet: > The original idea was that for docs / demos it would be nice to be able to show the info sent to the contract (i.e. the data visible to the blockchain

I see, I haven't ran this command (so don't know how it "looks"). I'll play with it shortly.

client/zeth/cli/zeth_mix.py
dtebbs: "mixer parameters" seems pretty unambiguous to me (i.e. "parameters to the mixer"), but we can reword it if you prefer.  "mix call parameters" alone could be slightly ambiguous, simply because we have client-side functions and commands called `mix`.

AntoineRondelet: This is a detail so I'm happy to keep "mixer params". For the record "mixers params" confuses me because: The Mixer is the contract name, hence "mixer parameters" makes me think about the parameters of its constructor and/or the value of the Mixer contract attributes (i.e. written in the Mixer contract's state). Wihtout looking at the code and just by the description of the flag I though I did a call to "get" the attributes values on the mixer to display them to the user here (i.e. Merkle tree depth, etc etc).

AntoineRondelet:
LGTM

PR #266, opened 2020-09-02
Fix mpc code

AntoineRondelet:
Fixes/Adds:
- Flag documentation and naming consistency
- Path to wrapped binaries
- Few documentation hiccups (small rephrasings and flag/args fixes for some commands)
- Streaming for challenges (for high degree QAPs)
- Change all "times" to UTC


mpc/README.md
```diff
@@ -81,9 +84,10 @@ Digest written to: response.bin.digest
 ...
 ```
 
-For phase2:
+For phase2 (assuming `mpc-client-phase2` is in the `PATH`):
```
dtebbs, line 87: I'm guessing that you hit an issue when the `mpc-coord-phase2` tool was not available?

Rather than users adding this to their path, it might be better for UX to update the code that looks for the tool in the default location (e.g. check for either `mpc-coord-phase2` or `mpc-client-phase2`, or use a flag to set client/server). This will avoid the user having to update his local environment, in most cases.

mpc/README.md
```diff
@@ -81,9 +84,10 @@ Digest written to: response.bin.digest
...
 ```console
-(env) $ phase2_contribute https://<host>[:<port>] contributor.key
+(env) $ phase2_contribute https://<host>[:<port>] contributor.key \
+            --mpc-tool $(which mpc-client-phase2)
```
dtebbs, line 90: (related to the above, we should be able to remove the need for this flag in most cases)

mpc/coordinator/powersoftau_command.py
```diff
@@ -81,4 +80,9 @@ def _exec(
 
 
 def _default_powersoftau_path() -> str:
-    return join(dirname(__file__), "..", "..", "..", "powersoftau")
+    """
+    Return the default path to the PoT binaries directory
+    """
+    return join(
+        dirname(__file__),
```
dtebbs, line 87: Did you mean to introduce this break?  All parameters fit on a single line (or if you prefer to break, each parameter should be on its own line to be consistent)

mpc/coordinator/powersoftau_process_command.py
```diff
@@ -45,4 +45,6 @@ def _exec(self, args: List[str]) -> bool:
 
 def _default_tool() -> str:
     from os.path import join, dirname
-    return join(dirname(__file__), "..", "..", "build", "src", "pot-process")
+    return join(
+        dirname(__file__), "..", "..",
+        "build", "mpc_tools", "pot-process")
```
dtebbs, line 50: Similar to the above, was this intentional? We usually have all params on one line, or one-per-line.

dtebbs (changes requested):
Just a few small comments.

mpc/coordinator/powersoftau_command.py
AntoineRondelet: Oh don't know why I broke the line here, you're right this is useless and inconsistent. Will move that up to the same line (same for below)

mpc/README.md
AntoineRondelet: The current code looks into default locations that are relative paths from the "contributor"'s repo. This makes the assumptions that:
1. the contributor compiled the binaries from source (which can be desirable actually - to verify the code the ends up being ran), and
2. has cloned `powersoftau` and `zeth` under the same parent directory. Any other structure will break since the tool will look in the wrong location.

If we want to package everything in docker images, and use a multi-stage build to build the binaries and move them in the final image, then we will get rid of all the useless code and just move binaries around, which won't be under `zeth/build/mpc_tools/mpc_phase2/`, same for PoT binaries which won't be under `/powersoftau/target/release`, hence the flag is quite helpful here.
Actually anyone running the `setup_env.sh` script will already have the Zeth binaries in is PATH, and thus should be able to run this command without any trouble.

I can remove the `--mpc-tool $(which mpc-client-phase2)` here and put `--mpc-tool <path-to-mpc-binary>`, that works for me either way - the second alternative just forces the user to look for the path to the MPC binary and paste it here.
Else, we need to modify:
`(Phase1 only) clone and build: https://github.com/clearmatics/powersoftau` to rather say: `(Phase1 only) clone and build: https://github.com/clearmatics/powersoftau under the same parent directory as Zeth` to make sure the default path in `powersofau_command.py` (i.e. `../../../powersoftau/target/<CONFIG>`) resolves to the PoT binaries which is not enforced here.

mpc/README.md
```diff
@@ -156,7 +160,7 @@ specifying properties of the MPC:
 {
     "server": {
         "contributors_file": "contributors.json",
-        "start_time": "2019-10-02 17:00:00",   # Time (server-local)
+        "start_time": "2019-10-02 17:00:00",   # Time in UTC (server-local)
```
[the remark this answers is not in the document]
AntoineRondelet: Yes ofc...! I meant `UTC` here, not local. Will remove the `server-local`...

I think it makes sense to just consider all times to be UTC, the MPC is a global process (the coordinator can be in Central Europe - CEST) while a participant can be in the West coast of the US. UTC is the "common denominator", and does not require any time conversions, so it just looks to be the most suitable choice to handle time. Commands like `timedatectl` gives you the time UTC, so you just need to `timedatectl` and paste the UTC result in the server config file before you run the coordinator and start the server.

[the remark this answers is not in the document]
AntoineRondelet: Right, switching to UTC requires to use https://docs.python.org/3/library/time.html#time.gmtime instead of `localtime`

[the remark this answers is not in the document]
dtebbs: > I think it makes sense to just consider all times to be UTC, the MPC is a global process (the coordinator can be in Central Europe - CEST) while a participant can be in the West coast of the US. UTC is the "common denominator", and does not require any time conversions, so it just looks to be the most suitable choice to handle time. Commands like timedatectl gives you the time UTC, so you just need to timedatectl and paste the UTC result in the server config file before you run the coordinator and start the server.

Yes, UTC is definitely the common zone and there are tools to convert.  My point was more that it might be easy for the coordinator admin to forget this step, and it can be very confusing to debug when this happens.  (So by requiring the zone to be specified we give a strict check against this - it's also a trivial code change.)

UTC-only will work of course, if you prefer that, but let's be careful to make it as clear as possible - including in the sample config files as well as this README.  I'd even consider changing the JSON attribute `start_time` to `start_time_utc` to make it absolutely clear, but perhaps that's overkill.

mpc/README.md
dtebbs: > The current code looks into default locations that are relative paths from the "contributor"'s repo. This makes the assumptions that:
> 
>     1. the contributor compiled the binaries from source (which can be desirable actually - to verify the code the ends up being ran), and
> 
>     2. has cloned `powersoftau` and `zeth` under the same parent directory. Any other structure will break since the tool will look in the wrong location.

Yes, exactly.  `powersoftau` might be a slightly different case since it's external, but for the `--mpc-tool` flag here, I'd expect binary to be in a "known" location if the user is using defaults.  If the user is doing something custom, that's exactly what the flag is for.

> If we want to package everything in docker images, and use a multi-stage build to build the binaries and move them in the final image, then we will get rid of all the useless code and just move binaries around, which won't be under `zeth/build/mpc_tools/mpc_phase2/`, same for PoT binaries which won't be under `/powersoftau/target/release`, hence the flag is quite helpful here.
> Actually anyone running the `setup_env.sh` script will already have the Zeth binaries in is PATH, and thus should be able to run this command without any trouble.

Ah, good point.  I usually keep a plain env so I forgot about the `setup_env.sh` script.  So that's true that the command will be in the path if it has been built locally in the default location (your assumption 1 above) and the user has run `setup_env.py`.  However, if setup_env.sh *has* been run, it seems like even more reason not to require the user to specify the location again here (if I understood correctly, that's part of the intention behind `setup_env.sh`).

In the case of a docker image, I'd expect the binaries to be placed somewhere convenient so that again the user doesn't have to give this flag (or we'd wrap everything in a script).  

> I can remove the `--mpc-tool $(which mpc-client-phase2)` here and put `--mpc-tool <path-to-mpc-binary>`, that works for me either way - the second alternative just forces the user to look for the path to the MPC binary and paste it here.
> Else, we need to modify:
> `(Phase1 only) clone and build: https://github.com/clearmatics/powersoftau` to rather say: `(Phase1 only) clone and build: https://github.com/clearmatics/powersoftau under the same parent directory as Zeth` to make sure the default path in `powersofau_command.py` (i.e. `../../../powersoftau/target/<CONFIG>`) resolves to the PoT binaries which is not enforced here.

I was thinking more of the mpc case, since we have more control over that, but in general I tend to aim for an approach where all commands are as simple as possible if the user has followed the README.  If the user has done something custom, we add hints such as: 
```
note: if mpc tools have been built or installed into a directory other than `build` in the repo root, the location must be specified with the `--mpc-tool` flag.  See `--help` for more details.
```

So, for the `powersoftau` location (which is a definitely more painful), your suggestion seems sensible to me: give a recommended install location (maybe with a note why), and "otherwise the location can be specified manually ... see --help".

(Eventually, we might consider making pwoersoftau a submodule, and have the python wrapper build it on demand, but that's far out of scope here, of course.)

mpc/README.md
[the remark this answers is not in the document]
AntoineRondelet: Yes that a good point.

> I'd even consider changing the JSON attribute start_time to start_time_utc to make it absolutely clear, but perhaps that's overkill.

I think this is valid, I like the idea. It's more explicit and cannot hurt.

> UTC-only will work of course, if you prefer that

I would favor UTC only here indeed. It is easier to use UTC everywhere and easier to extend the current code base to have a better UX.
E.g. For now we do send an email to the contributor when his turn starts. A more user friendly piece of code may want - at the coordinator server's start time - to send an email to ALL contributors with an estimated starting time for their contribution window (so that they do not need to wait for their emails and can plan when - roughly - they'll need to be attentive, i.e. when they need to run `phaseX_contribute --wait-interval XX`). This can be done by carrying out some benchmark for all contribution steps (fetching challenge, computing, uploading response + coordinator's work between contributions -> all these steps are bandwidth/CPU dependent but benchmarks can be used to roughly know how big the contribution window needs to be). Afterwards, you multiply by the contributor's position in the pool + do some basic date arithmetic and you can send the approx contribution time in UTC to everyone up front.
That's just an example of additional step that can be done, and ofc you can do it using the server's time zone and then convert all results in UTC, but this is extra code that shouldn't need to be written IMO.

> My point was more that it might be easy for the coordinator admin to forget this step, and it can be very confusing to debug when this happens.

Indeed, I (almost) always use UTC by default - so I'd be confused if I needed to use the local time zone. Others may be confused the other way around. Good documentation + small changes in the code as you suggested (i.e. `start_time` -> `start_time_utc`) should be enough to carry the message out (and tbh I am yet to be convinced why a sysadmin managing a server communicating with other machines in arbitrary time zones would decide to use the local time zone on his server - even for logging activities - but I am not a sysadmin, so I may well be very wrong here!)

mpc/README.md
AntoineRondelet: >  However, if setup_env.sh has been run, it seems like even more reason not to require the user to specify the location again here (if I understood correctly, that's part of the intention behind setup_env.sh).

Exact, that's why I used `$(which CMD)` as a way to directly get the path to the command (and not requiring the user to specify it).

> In the case of a docker image, I'd expect the binaries to be placed somewhere convenient

Yes I moved all the binaries in the PATH by default (see: Docker PR for more details). So users can copy and paste:
```
phase2_contribute https://mpc-contributor:<port> contributor.key \
            --mpc-tool $(which mpc-client-phase2)
```
and it will just work.

> so that again the user doesn't have to give this flag

How would it be possible with the current code base? The default path is `../../build/mpc_tools/mpc_phase2/<tool>`, so to save the flag from the command, one would need to build nested folders that reflect the relative path above (same for PoT btw), no?

AntoineRondelet: > Ah, good point. I usually keep a plain env so I forgot about the setup_env.sh script. So that's true that the command will be in the path if it has been built locally in the default location (your assumption 1 above) and the user has run setup_env.py.

Note that instead of modifying the PATH, we can also move the binaries to `/usr/local/bin` (move them in the PATH) with a `make install` rule. As long as they are in the PATH, we can locate them with `which`. 
We could actually "wrap" the `which` command to locate the binaries, didn't aim to change too many things here but that would remove the need for the flag (which is probably what you had in mind?)..

AntoineRondelet: As discussed off-line the best approach seems to be to mimic what we've done on the cpp side, where we use different levels of priority between flags, config files and env vars. Likewise, we can use env vars here to point to the binaries, and fall back to the path that is currently hardcoded in the code if the env var is not defined.

dtebbs: Just to complete the discussion, it sounds like the original issue was with trying to manually move binaries out of the build directory.
This case has not been addressed (the default build does not support this and the assumption was that users would leave the binaries in the default build location), and hence the `--mpc-tool` flag was required in these circumstances.

The right way to address this would be to create some kind of install process, so that the destination is specified once and all tools can be installed with the information they need to find each other.

For the ad-hoc case of the binaries in arbitrary locations, an env-var can be used as an alternative to this flag.  This would still rely on several assumptions (like the binaries being statically linked or being able to find their dynamic dependencies), but could be useful for Docker contrainers where the state of the system can be preconfigured.

But ideally, for deployment, we would support:
```
$ cmake [CMAKE_INSTALL_PREFIX=/usr/local]
$ make install
```

Since the tools are generally only intended to be used via the python wrappers, it could be really nice to encapsulate everything in a python pacakge (so the native tools are build at `pip install` time, as with other packages).  This would give uses a one-shot install command.

mpc/coordinator/powersoftau_process_command.py
AntoineRondelet: Fixed in https://github.com/clearmatics/zeth/pull/266/commits/98bedfa67af3f884f53373ec5a403ffdf737bccc

mpc/coordinator/powersoftau_command.py
AntoineRondelet: Fixed in https://github.com/clearmatics/zeth/pull/266/commits/98bedfa67af3f884f53373ec5a403ffdf737bccc

mpc/README.md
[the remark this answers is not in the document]
AntoineRondelet: #268 changes the time units to UTC only

dtebbs (approved):

PR #267, opened 2020-09-04
SSL support in client

dtebbs:


client/zeth/core/utils.py
```diff
@@ -13,26 +13,33 @@
...
+        url: str,
+        certificate: Optional[str] = None,
+        insecure: bool = False) -> Any:
     """
     Create a Web3 context from an http URL.
     """
-    return Web3(HTTPProvider(
-        url,
-        request_kwargs={'timeout': WEB3_HTTP_PROVIDER_TIMEOUT_SEC}))
+    assert not certificate or exists(certificate)
```
AntoineRondelet, line 36: Why did you `assert` here instead of raising a `FileNotFoundError` for eg?

dtebbs: switched this check to `raise`

AntoineRondelet:
Thanks, LGTM

PR #268, opened 2020-09-08
Switched coordinator to use UTC time only

AntoineRondelet:
Used the appropriate functions to manipulate UTC times and added check to constraint UTC/GMT times in the configuration file.


mpc/coordinator/server_configuration.py
```diff
@@ -5,14 +5,15 @@
...
+import calendar
 import json
 import time
 from os.path import dirname, exists, join
 from typing import Dict, cast, Optional
 
 JsonDict = Dict[str, object]
 
-TIME_FORMAT = "%Y-%m-%d %H:%M:%S"
+TIME_FORMAT = "%Y-%m-%d %H:%M:%S UTC"
```
dtebbs, line 16: You might consider using `"%Y-%m-%d %H:%M:%S %Z"` here, and keeping `time.mktime` below, instead of `calendar.timegm`.  That way, the timezone information then gets included in the time structure, and is used when converting to POSIX seconds.  (This format will correctly extract the UTC zone from the time struct created by `time.gmtime`, so the output will still be correct.)

mpc/coordinator/server_configuration.py
```diff
@@ -88,7 +89,7 @@ def from_json(config_json: str) -> Configuration:
         return Configuration._from_json_dict(json.loads(config_json))
 
     def _to_json_dict(self) -> JsonDict:
-        start_local = time.localtime(self.start_time)
+        start_local = time.gmtime(self.start_time)
```
dtebbs, line 92: `start_local` -> `start_time`?

mpc/coordinator/server_configuration.py
```diff
@@ -102,7 +103,7 @@ def _to_json_dict(self) -> JsonDict:
         }
 
     def _to_json_template_dict(self) -> JsonDict:
-        start_local = time.localtime(self.start_time)
+        start_local = time.gmtime(self.start_time)
```
dtebbs, line 106: `start_local` -> `start_time`?

mpc/coordinator/server_configuration.py
```diff
@@ -131,7 +132,7 @@ def _from_json_dict(
             email_password_file = join(dirname(config_path), email_password_file)
         return Configuration(
             cast(str, json_dict["contributors_file"]),
-            time.mktime(start_local),
+            calendar.timegm(start_local),
```
dtebbs, line 135: `start_local` -> `start_time`

dtebbs (changes requested):

mpc/coordinator/server_configuration.py
AntoineRondelet: Switching to `UTC` instead of `%Z` is done on purpose (see https://github.com/clearmatics/zeth/pull/268/commits/210d8eaa3fbe99d6b347af50819228b1e7364b91)
If time needs to be given in UTC in the config, why would we want to accept all times of the form `%Y-%m-%d %H:%M:%S %Z`? Here parsing the configuration will fail if time isn't UTC in the config, which is a good thing - no time conversion, no surprise expected, it's all UTC.

dtebbs: OK.  It's just small usability thing, so let's leave it like this and move on.  Just some comments for future reference in case this gets revisited:

- as long as zone is explicit there should be no room for error (a mistake in zone is similar to a mistake in other field such as hous or day).

- however, there is no massive issue in locking to UTC, but in this case it's a little confusing that the timezone *must* be specified but it's *also redundant* (can only every be one value).  Users seeing it there, users will naturally expect to be able to put any valid timezone.  Therefore, if we come back to this and still decide to lock in to UTC (and as a general usability principle) I would probably favour renaming the attribute to `start_time_utc` and not requiring the user to specify it.

AntoineRondelet: > however, there is no massive issue in locking to UTC, but in this case it's a little confusing that the timezone must be specified but it's also redundant (can only every be one value). Users seeing it there will naturally expect to be able to put any valid timezone. Therefore, if we come back to this and still decide to lock in to UTC (and as a general usability principle) I would probably favour renaming the attribute to start_time_utc and not requiring the user to specify it

Adding UTC as part of the format is to force the user to specify it (this redundancy is on purpose here). Happy to switch to `start_time_utc`, that's achieving the same thing (removes the 3 characters `UTC` from what the user needs to write).

>  It's just small usability thing,

I think all these discussion on usability are very good and very important, but let's bear in mind that we are talking about MPC code here. The MPC should (well.... in theory - if no changes in the circuit are made later on) be carried out **only once** (Groth16 is not universal and updatable). As such, this code should normally be ran only once for each phase. Additionally, not everyone is expected to be the coordinator. So yes usability is very important, but an extra flag or extra few characters to be written in the coordinator config do not look as big of a deal to me as if it was the wallet code (in which case UX needs to be as good as it can be because the wallet is used frequently by users, and, even more importantly, bad UX for the wallet can translate is unexpected/risky/leaky behavior from the user).
That being said, more than happy to see the code refactored and UX improved, but that would require (for the UX to be very nice) quite some refactor which is not he purpose of the currently opened PR.

AntoineRondelet: Moved to `start_time_utc` in https://github.com/clearmatics/zeth/pull/268/commits/c75809c77dad76cec45e24c5711078cfe7f8db0e

mpc/coordinator/server_configuration.py
AntoineRondelet: See: https://github.com/clearmatics/zeth/pull/268/commits/c75809c77dad76cec45e24c5711078cfe7f8db0e

testdata/mpc_phase1_server_config.json
```diff
@@ -3,9 +3,9 @@
         "_contributors_file_description": "Relative path of contributors file",
         "contributors_file": "../testdata/mpc_contributors.json",
 
-        "_start_time_description": "Time (local) that first contribution is allowed",
-        "_start_time_example": "2019-11-05 13:33:13",
-        "start_time": "TIME",
+        "_start_time_utc_description": "Time (local) that first contribution is allowed",
```
dtebbs, line 6:
```suggestion
        "_start_time_utc_description": "Time (in UTC) that first contribution is allowed",
```

dtebbs (changes requested):
Just one remaining mention of local time. Aside from that, LGTM.

testdata/mpc_phase1_server_config.json
AntoineRondelet: Thanks

dtebbs (approved):

Issue #258, opened 2020-08-28
Docker autobuild is constently failing


AntoineRondelet:
For some reason the docker autobuild of the Dockerfile-base on `develop` has been constantly failing.
I looked at the logs on DockerHub but cannot really pin point what needs to be done to solve the issue and make `clearmatics/zeth-base` build correctly.
Building the image locally via `docker build --no-cache -f Dockerfile-base -t clearmatics/zeth-base .` works well (I passed the `--no-cache` flag to be sure that all steps were ran from scratch and catch any potential issue with the image).

Could I receive some assistance on this one?
Any help would be appreciated @eastata @Klazomenai :)

eastata:
https://hub.docker.com/repository/registry-1.docker.io/clearmatics/zeth-base/builds/5525d7a3-8319-4fa6-a713-78b11fa4a8c2
```
virtual memory exhausted: Out of memory
```

AntoineRondelet:
Thanks @eastata

Klazomenai:
@eastata Is this issue a sub-task of #278 which will be covered within that issue and can be closed?

eastata:
Should be resolved bt PR https://github.com/clearmatics/zeth/pull/313

PR #273, opened 2020-09-10
Sigloi: v0.5

AntoineRondelet:
Merging the content of v0.5 to master


scripts/test_zeth_cli_token
```diff
@@ -3,6 +3,13 @@
 set -e
 set -x
 
+if ! [ "$1" == "" ] ; then
+    echo == RUNING ON NETWORK: $1 ==
```
AntoineRondelet, line 7: RUNNING (I'd also add surrounding `"`)

AntoineRondelet:
This failing docker check is really annoying. I raised it in https://github.com/clearmatics/zeth/issues/258 already, but we'd need to fix it sooner rather then later because this is very annoying... I cloned a fresh version of the repo and built the docker images, then pushed them to the Hub manually so this is not a blocker for this PR to go through, but continuously having failing checks (especially on new "releases" PR) is very annoying.

CMakeLists.txt
```diff
@@ -269,26 +290,33 @@ add_compile_options(-Wall -Wextra)
...
+  # Import configuration to compile targets with sanitizers
+  include(cmake/sanitizers.cmake)
 
-# Import configuration to generate the code documentation if option is set
-if("${GEN_DOC}")
-    include(cmake/documentation.cmake)
-endif()
+  # Import configuration to generate the code documentation if option is set
+  if("${GEN_DOC}")
+      include(cmake/documentation.cmake)
```
dtebbs, line 303: indentation

dtebbs:
> I cloned a fresh version of the repo and built the docker images, then pushed them to the Hub manually so this is not a blocker for this PR to go through

So is this expected to pass once when this is merged to master?
If not, can this be removed from the checks until it passes?

> but continuously having failing checks (especially on new "releases" PR) is very annoying.

Yes indeed.  I get a 404 when I try to see the details, so not sure whether it's something trivial or likely to keep happening.

dtebbs (approved):

dtebbs:
LGTM when #279 is merged.

AntoineRondelet:
> So is this expected to pass once when this is merged to master?
> If not, can this be removed from the checks until it passes?

I don't think so. I get a 404 when trying to figure out what's happening, and can't see the logs when I try to access them from the Hub directly, so tricky to debug and fix from this perspective...
We'd need support from DevOps to fix this (as per my issues #258 and #278) as it seems this auto build was configured with an account I don't have (and don't want) to have control over.

AntoineRondelet:
All the checks have successfully be ran to the exception of the docker autobuild as per my comment above and the issues raised.
All the docker images have been built with the `0.5` tag and successfully pushed to the Hub.
Merging this PR.

Issue #281, opened 2020-09-14
`estimateGas()` raises an error when RPCing a non-validator node


AntoineRondelet:
The `estimateGas()` function is relied upon in various functions in the `test_commands` and `zeth_helper` (e.g. `test_commands/deploy_test_token.py` (in `mint`) and `zeth_helper token-approve`) to set the `gas` of a transaction. Nevertheless, some unexpected behaviors arose when pointing to a non-validator RPC endpoint which for some reason triggered an error:
```
Traceback (most recent call last):
  File "../test_commands/deploy_test_token.py", line 123, in <module>
    deploy_test_token()  # pylint: disable=no-value-for-parameter
  File "$ZETH/client/env/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "$ZETH/client/env/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "$ZETH/client/env/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "$ZETH/client/env/lib/python3.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "../test_commands/deploy_test_token.py", line 54, in deploy_test_token
    EtherValue(mint_amount, 'ether'))
  File "../test_commands/deploy_test_token.py", line 119, in mint_token
    sender_eth_private_key=deployer_private_key)
  File "$ZETH/client/zeth/core/contracts.py", line 359, in send_contract_call
    transaction = call.buildTransaction(tx_desc)
  File "$ZETH/client/env/lib/python3.7/site-packages/web3/contract.py", line 1228, in buildTransaction
    **self.kwargs
  File "$ZETH/client/env/lib/python3.7/site-packages/web3/contract.py", line 1512, in build_transaction_for_function
    prepared_transaction = fill_transaction_defaults(web3, prepared_transaction)
  File "cytoolz/functoolz.pyx", line 232, in cytoolz.functoolz.curry.__call__
  File "$ZETH/client/env/lib/python3.7/site-packages/web3/utils/transactions.py", line 55, in fill_transaction_defaults
    default_val = default_getter(web3, transaction)
  File "$ZETH/client/env/lib/python3.7/site-packages/web3/utils/transactions.py", line 26, in <lambda>
    'gas': lambda web3, tx: web3.eth.estimateGas(tx),
[...]

PR #282, opened 2020-09-14
[WIP] Added missing gas argument to prevent error related to `estimateGas()`

AntoineRondelet:
Addresses: https://github.com/clearmatics/zeth/issues/281 by specifying gas for each transaction.


AntoineRondelet:
Closing this PR since the issue is related to the Autonity node rather than Zeth per say. As such, there is no desire to hardcode gas values here. Instead, Autonity must be fixed to make sure that the `estimateGas` function can be ran successfully.

PR #280, opened 2020-09-14
Custom ganache-cli, supporting operations on bw6-761 and bls12-377

dtebbs:
- use custom version of ganache-cli
- tests for new precompiled contracts exposed by custom ganache-cli

#285 addresses the use of a submodule here, rather than just the npm mechanisms

PR #283, opened 2020-09-15
Improved encoding of objects passed to web3

dtebbs:
We encode verification keys, proofs and input values into evm words.  This PR:
- improves the interface to this functionality
- makes the result more consistent over proofs / keys
- supports converting elements from pairings other than alt-bn128.

(Depends on #280)

Note, there is still room for stronger types around the handling of keys and proofs in python.
This should be addressed in a future PR.

### DIFF
--- a/client/tests/test_zksnark.py
+++ b/client/tests/test_zksnark.py
@@ -184,6 +184,64 @@ VERIFICATION_KEY_BW6_761_GROTH16_PARAMETERS = [
     ]
 ]
 
+VERIFICATION_KEY_ALT_BN128_GROTH16 = {
+    "alpha": [
+        "0x009d7309d79d5215384a7a9a1d9372af909582781f388a51cb833c87b8024519",
+        "0x012816ef6069ef1e40eaab0a111f9b98b276dbf2a3209d788eb8ce635ce92a29",
+    ],
+    "beta": [[
+        "0x017abb9470ccb0ef09676df87dbe181a9ed89ba1cf1e32a2031d308b4c11a84f",
+        "0x01774daba40ce4c9fe2d2c6d17a3821b31ec63a77ebea2dab8b3218fd7eb90f9",
+    ], [
+        "0x18ce3769d0c1e29aa799a5928b1c524a5a85326c4b16463530bfdcab82f55ef6",
+        "0x1a9234f3340fb85ae722ed052b8dcf63193c423791d9c43ab725a35286bda170",
+    ]],
+    "delta": [[
+        "0x19c19b1795e634573c0514de0cea5bd05d88c24b08aeadc03ec4686ee6741b80",
+        "0x01a00d16c4d2805e248debf48ea0771e627e2bfb95198df0cbe09a1eb4879fe5",
+    ], [
+        "0x00361ca07388d760898e0969f3b9a3d6d751b83d770007761e1c5cc798852ed8",
+        "0x009a7d27c8392eefe1ba23a52d509cda59ba3c5acc95765d1146a998c780277f",
+    ]],
+    "ABC": [[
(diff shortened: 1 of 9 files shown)


client/zeth/core/zksnark.py
```diff
@@ -158,17 +156,16 @@ def proof_to_proto(
         return extproof_proto
 
     @staticmethod
-    def mixer_proof_parameters(extproof: GenericProof) -> List[List[Any]]:
+    def proof_to_evm_parameters(extproof: GenericProof) -> List[List[int]]:
```
AntoineRondelet, line 159: `proof_to_evm_parameters` -> `proof_to_evm_words`?

AntoineRondelet: (same for functions below)

client/tests/test_zksnark.py
```diff
@@ -8,8 +8,321 @@
...
+        self.assertEqual(PROOF_BLS12_377_GROTH16_PARAMETERS, proof_parameters)
+
+    def test_bw6_761_groth16_verification_key_parameters(self) -> None:
+        vk = VERIFICATION_KEY_BW6_761_GROTH16
+        vk_parameters_expect = VERIFICATION_KEY_BW6_761_GROTH16_PARAMETERS
+        vk_parameters = \
+            zksnark.Groth16SnarkProvider().verification_key_to_evm_parameters(vk)
+        self.assertEqual(vk_parameters_expect, vk_parameters)
+
+    def test_bw6_761_groth16_proof_parameters(self) -> None:
```
AntoineRondelet, line 320: I can imagine that this test (BW6) is here for completeness as `libzeth` should be generic enough and re-used, especially in Zecale, where BW6 is relevant (no Zeth proof will be generated over BW6). As such, I'd add a test case for the other supported curve, namely `bn254`, for completeness, even if we know that the elements of the associated fields (base/scalar) fit in a single EVM word.

client/zeth/core/zksnark.py
dtebbs: I'm not attached to this encoding name, but couldn't think of a better one.

These fucntions generate the set of parameters to pass to the contract call.  By coincidence it happens to be all evm_words here, but as we move to use more structs the types here could change (in fact, a later PR flattens this to `List[int]`, but I expect it will eventually be deeper and we may want to use `bytes` etc.

I considered something like `eth_abi`, but that's the result of later encoding this to a binary string.  Other options could be `contract_parameters` or `web3_parameters` or something?

client/tests/test_zksnark.py
dtebbs: Added tests for alt_bn128.

client/zeth/core/zksnark.py
AntoineRondelet: > These fucntions generate the set of parameters to pass to the contract call.
> Other options could be contract_parameters or web3_parameters or something?

Why not naming the function: `proof_obj_to_contract_parameters` then? (same for other related functions)
The EVM is a stack-based VM processing (popping/pushing) 256-bit long words. Hence, if the goal is to convert objects (like `extended_proof`) to contract parameters, I find the current name slightly misleading, as I'd expect everything to be systematically converted to 256-bit long words instead

client/tests/test_zksnark.py
AntoineRondelet: Thanks

client/zeth/core/zksnark.py
dtebbs: > > Other options could be contract_parameters or web3_parameters or something?

> Why not naming the function: proof_obj_to_contract_parameters then?

Renamed the encoding to `contract_parameters`

AntoineRondelet: Thanks @dtebbs

AntoineRondelet:
LGTM

PR #284, opened 2020-09-16
CI tidy up

dtebbs:
- Fix for ci command that is permitted to fail on some platforms
- Enable client unit tests running on CI host
- Update some python packages to take advantage of prebuilt wheels
- Remove some duplication in the CI actions


AntoineRondelet (commented):
Thanks @dtebbs

Quick q:
1. You mention `Update some python packages to take advantage of prebuilt wheels` -> Beyond bumping version numbers, what do you do to "take advantage of prebuilt wheels"?
2. `Fix for ci command that is permitted to fail on some platforms` -> What did you do here to allow failures on some platforms? Do you refer to the fact that the `onpullrequest.yml` config only contains a check for linux?

.github/workflows/onpullrequest.yml
```diff
@@ -0,0 +1,19 @@
+# Actions for pull requests only
+name: zeth-ci-pull-request
+
+on:
+  pull_request:
+
+env:
+  MACOS_BREW_PACKAGES: "pkg-config libomp"
```
AntoineRondelet, line 8: Do you need this env var?

dtebbs:
> Beyond bumping version numbers, what do you do to "take advantage of prebuilt wheels"?

I don't think I did anything else.  The builds were building grpcio-tools because it was at 1.24 and there were no precompiled wheels for that version.

> Fix for ci command that is permitted to fail on some platforms -> What did you do here to allow failures on some platforms? Do you refer to the fact that the onpullrequest.yml config only contains a check for linux?

This is the `|| echo` added in the `ci` script (already used for a few other commands).  For some reason this suddenly started failing on the CI machines.

.github/workflows/onpullrequest.yml
dtebbs: removed this env var

AntoineRondelet:
> > Beyond bumping version numbers, what do you do to "take advantage of prebuilt wheels"?
> 
> I don't think I did anything else. The builds were building grpcio-tools because it was at 1.24 and there were no precompiled wheels for that version.
> 
> > Fix for ci command that is permitted to fail on some platforms -> What did you do here to allow failures on some platforms? Do you refer to the fact that the onpullrequest.yml config only contains a check for linux?
> 
> This is the `|| echo` added in the `ci` script (already used for a few other commands). For some reason this suddenly started failing on the CI machines. (i.e. it's separate from the other changes, but required to prevent the macos build failing).

I see thanks.

AntoineRondelet:
LGTM

PR #288, opened 2020-09-21
Compact groth16 arguments

dtebbs:
Compact the proof and vk into single arrays of evm words.  Saves gas and reduces the number of parameters required in applictions where vk and proof must be passed as arguments.

Depends on #287


client/zeth/core/contracts.py
```diff
@@ -220,14 +220,13 @@ def mix_parameters_as_contract_arguments(
     Convert MixParameters to a list of eth ABI objects which can be passed to
     the contract's mix method.
     """
-    proof_params: List[Any] = zksnark.proof_to_contract_parameters(
-        mix_parameters.extended_proof)
-    proof_params.extend([
+    proof_params: List[Any] = [
```
AntoineRondelet, line 223: What about renaming this one as `mix_params`?

zeth_contracts/contracts/Groth16Mixer.sol
```diff
@@ -41,35 +41,42 @@ contract Groth16Mixer is BaseMixer {
 
     VerifyingKey verifyKey;
 
-    // Constructor
+    // Constructor.  Form of vk is:
+    //    uint256[2] Alpha,
+    //    uint256[4] Beta,
+    //    uint256[4] Delta,
+    //    uint256[] ABC_coords
```
AntoineRondelet, line 48: Side note: Now that we support several curves (i.e. bn/bls) may be worth to clarify these comments. This contract is tailored for bn254, and won't work with any other groups. So May be worth specifying under with circumstances your comment above holds (i.e. `ALT_BN128` set in cpp  code config) or simply rephrase to make this generic, e.g.:
```
// Constructor. The form of vk is:
//     alpha = vk[:2*N]
//     beta = vk[2*N:6*N]
//     delta = vk[6*N:10*N]
//     ABC_coords = vk[10*N:]
//  where:
//     - q is the characteristic of the base field of the pairing group used
//     - N = 2*\ceil{bitLen(q)/256} (# EVM words necessary to represent elements of Fq)
```

AntoineRondelet: Side note2: Making the constructor below generic shouldn't be really hard, but troubles may start with functions below, so that may be better to have different contracts to avoid any complexity related to generalization (and so my comment above may not apply anymore if the group name appears in the contract name)

zeth_contracts/contracts/Pghr13Mixer.sol
```diff
@@ -230,48 +225,39 @@ contract Pghr13Mixer is BaseMixer {
...
     function verifyTx(
-        uint256[2] memory a,
-        uint256[2] memory a_p,
-        uint256[4] memory b,
-        uint256[2] memory b_p,
-        uint256[2] memory c,
-        uint256[2] memory c_p,
-        uint256[2] memory h,
-        uint256[2] memory k,
+        uint256[18] memory proof_data,
```
AntoineRondelet, line 228: Name is inconsistent with argument name in `verifyTx` for Groth16

dtebbs: Renamed the groth16 version from `proof` to `proof_data`. (In the pghr13 case this function fills out an actual Proof structure, which is naturally named `proof`.  The groth16 version is optimized to pull directy from the memory buffer.)

client/zeth/core/contracts.py
dtebbs: renaed to `mix_params`

AntoineRondelet:
LGTM

PR #289, opened 2020-09-28
Client proof types

dtebbs:
Stricter types for generic Proof and VerificationKey types and snark API refactor.
(Allows more versatility in data structures that use proofs, verification keys and extended proofs)

### DIFF
--- a/client/test_commands/scenario.py
+++ b/client/test_commands/scenario.py
@@ -237,8 +237,8 @@ def charlie_double_withdraw(
     assert attack_primary_input4 != 0
 
     print("proof = ", proof)
-    print("proof[inputs][3] = ", proof.inputs[3])
-    print("proof[inputs][4] = ", proof.inputs[4])
+    print("proof.inputs[3] = ", proof.inputs[3])
+    print("proof.inputs[4] = ", proof.inputs[4])
     proof.inputs[3] = hex(attack_primary_input3)
     proof.inputs[4] = hex(attack_primary_input4)
     # ### ATTACK BLOCK
--- a/client/test_commands/scenario.py
+++ b/client/test_commands/scenario.py
@@ -7,6 +7,7 @@
 from zeth.core.mixer_client import MixerClient, OwnershipKeyPair, \
     joinsplit_sign, encrypt_notes, get_dummy_input_and_address, \
     compute_h_sig, JoinsplitSigVerificationKey
+from zeth.core.zksnark import IZKSnarkProvider
 import zeth.core.contracts as contracts
 from zeth.core.constants import ZETH_PRIME, FIELD_CAPACITY
 import zeth.core.signing as signing
@@ -143,6 +144,7 @@ def charlie_withdraw(
 
 
 def charlie_double_withdraw(
+        zksnark: IZKSnarkProvider,
         zeth_client: MixerClient,
         mk_tree: MerkleTree,
         input1: Tuple[int, ZethNote],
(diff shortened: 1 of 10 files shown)


client/test_commands/scenario.py
```diff
@@ -234,11 +236,11 @@ def compute_h_sig_attack_nf(
     assert attack_primary_input3 != 0
     assert attack_primary_input4 != 0
 
-    print("proof_json => ", proof_json)
-    print("proof_json[inputs][3] => ", proof_json["inputs"][3])
-    print("proof_json[inputs][4] => ", proof_json["inputs"][4])
-    proof_json["inputs"][3] = hex(attack_primary_input3)
-    proof_json["inputs"][4] = hex(attack_primary_input4)
+    print("proof => ", proof)
+    print("proof[inputs][3] => ", proof.inputs[3])
```
AntoineRondelet, line 240: Why not: `print("proof.inputs[3] = ", proof.inputs[3])`? (same for the line below)

client/tests/test_zksnark.py
```diff
@@ -2,18 +2,18 @@
...
 from unittest import TestCase
 
 
 # pylint: disable=line-too-long
-VERIFICATION_KEY_BLS12_377_GROTH16 = {
+VERIFICATION_KEY_BLS12_377_GROTH16 = Groth16.VerificationKey.from_json_dict({
     "alpha": [
         "0x009d7309d79d5215384a7a9a1d9372af909582781f388a51cb833c87b8024519cf5b343cb35d49a5aa52940f14b7b8e7",  # noqa
-        "0x012816ef6069ef1e40eaab0a111f9b98b276dbf2a3209d788eb8ce635ce92a29c2bcdaa3bb9b375a8d3ee4325c07f4ea"  # noqa
+        "0x012816ef6069ef1e40eaab0a111f9b98b276dbf2a3209d788eb8ce635ce92a29c2bcdaa3bb9b375a8d3ee4325c07f4ea",  # noqa
```
AntoineRondelet, line 16: Why did you need to add this comma here?

client/test_commands/scenario.py
dtebbs: The original code predates me, so not sure why it was `=>` instead of `=`.  Changed to `=`.

client/tests/test_zksnark.py
dtebbs: It's not needed as such.  Many of our lists and dictionaries have a trailing comma, so I added it to match those.

client/test_commands/scenario.py
AntoineRondelet: Oh yes, there were two things in my comment above:
- `=>` -> `=`
- `proof.[inputs][3]` -> `proof.inputs[3]`

client/tests/test_zksnark.py
AntoineRondelet: Adding this comma here contrasts with the rest of the block actually: https://github.com/clearmatics/zeth/blob/client-proof-types/client/tests/test_zksnark.py#L13-L39
which itself contrasts with: https://github.com/clearmatics/zeth/blob/client-proof-types/client/tests/test_zksnark.py#L172-L198 (where you add final commas everywhere)

client/test_commands/scenario.py
dtebbs: updated

client/tests/test_zksnark.py
dtebbs: updated.  (yes, there are examples of each - some data was copy-pasted from json)

AntoineRondelet:
LGTM

PR #300, opened 2020-10-15
WIP: Mimc with e=11

dtebbs:
(Marking WIP for now.  Looking into restrictions on valid exponents.)


AntoineRondelet:
I would be keen to have these changes documented in the specs in parallel to such implementation, see: https://github.com/clearmatics/zeth-specifications/issues/1

dtebbs:
These changes are now obsolete (the latest code supports `exponent` and `round` parameters, and 11 is not known to be a secure exponent.

PR #305, opened 2020-10-28
Fix instructions to run custom ganache

AntoineRondelet:
There were a few annoying traps which required more documentation.
- The version numbers need to be more strict on `node` and `nvm` since the FFI library used does not seem to support node versions > 10 (see: https://github.com/node-ffi/node-ffi#requirements). Furthermore, I got some errors related to dependency syntax (aliases) which according to https://stackoverflow.com/questions/54085943/npm-err-invalid-dependency-type-requested-alias requires a version of `npm` > 6.9 (and not "at least 6.4.1"). Running `nvm use 10 && nvm install-latest-npm` allows to use the right versions. Always making sure to provide a minimal working Dockerfile allows to catch these version mismatch and is a good way to save time going forward
- The `npm install` carried out on the CI server (see here: https://github.com/clearmatics/zeth/blob/develop/scripts/ci#L27) does not seem to be executed (see: https://github.com/clearmatics/zeth/runs/1265491500#step:3:490 for instance) - seems like `npm install --unsafe-perm` fixes this.

For some reason the `dist` doesn't get built for our custom `ethereumjs-vm` package as mentioned in more details in this ticket: https://github.com/clearmatics/ganache-cli/pull/1#issue-511485994

### DIFF
--- a/.github/workflows/onpullrequest.yml
+++ b/.github/workflows/onpullrequest.yml
@@ -12,5 +12,8 @@ jobs:
     - uses: actions/checkout@v1
       with:
         submodules: recursive
+    - uses: actions/setup-python@v2
+      with:
+        python-version: '3.8'
     - name: Execute
       run: CI_EVENT_NAME=pull_request CI_USE_DOCKER=1 CI_CONFIG=Release scripts/ci build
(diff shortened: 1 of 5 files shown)


dtebbs (approved):
LGTM

PR #309, opened 2020-11-06
Mimc configurations (depends on #312)

dtebbs:
Supercedes #300.
Refactor to allow esy overriding of hash in merkle tree.
Add support for MiMC31 using the BLS12-377 scalar field modulus, where 31 = smallest e s.t. e= 2^1-1 and gdc(e, r-1) == 1. 
Not enabled by default.

Documented in clearmatics/zeth-specifications#4.


client/test_commands/test_mimc_contract.py
```diff
@@ -0,0 +1,60 @@
...
+def test_mimc31(instance: Any) -> None:
+    x = int(28948022309329048855892746252171976963317496166410141009864396001978282409983)  # noqa
+    y = int(14220067918847996031108144435763672811050758065945364308986253046354060608451)  # noqa
+    h = MiMC31().hash(x, y).to_bytes(32, 'big')
+
+    result = instance.functions.test_mimc31(
+        x.to_bytes(32, 'big'), y.to_bytes(32, 'big')).call()
+
+    print(f"h={int.from_bytes(h, byteorder='big')}")
+    print(f"r={int.from_bytes(result, byteorder='big')}")
```
AntoineRondelet, line 33: The log and the variable names are not consistent here.
`print(f"r={int.from_bytes(result, byteorder='big')}")` -> `print(f"result={int.from_bytes(result, byteorder='big')}")`

client/tests/test_mimc.py
```diff
@@ -2,17 +2,40 @@
...
+        self.assertEqual(h, result)
+
+    def test_mimc31_round(self) -> None:
+        m = MiMC31()
+        x = 340282366920938463463374607431768211456
+        k = 3614637061043937583146271435827337369189798160947949526058695634226054692860  # noqa
+        c = 5775606169419625606859319496982126279674858730791300481051019590436651369410  # noqa
+        h = 5523634951166384704739554074217840169048851347397743343350526776025419511991  # noqa
+        result = m.mimc_round(x, k, c)
+        self.assertEqual(h, result)
```
AntoineRondelet, line 34: May be worth to avoid the use of one letter names for clarity (c = round_constant, k = key, m = mimc, x = message (msg), h =  expected_res)

client/zeth/core/mimc.py
```diff
@@ -2,53 +2,106 @@
...
-        seed = seed or self.seed
-        res = message % self.prime
+            ek: int) -> int:
+        result = message % self.prime
         key = ek % self.prime
+        round_constant: int = self.seed
 
-        # In the paper the first round constant is set to 0
-        res = self.mimc_round(res, key, 0)
+        # The rc in round 0 is 0 (see [AGRRT16])
```
AntoineRondelet, line 44: rc -> round constant

client/zeth/core/mimc.py
```diff
@@ -2,53 +2,106 @@
...
+            91)
+
+    def mimc_round(self, message: int, key: int, rc: int) -> int:
+        xored = (message + key + rc) % self.prime
+        return xored ** 7 % self.prime
+
+
+class MiMC31(MiMCBase):
+    """
+    MiMC implementation using exponent of 11 and 51 rounds. Note that this is
```
AntoineRondelet, line 87: Wrong exponent (`ceil(log(rBLS,2)/log(31,2)) = 51`). May be a residual from the switch away from MiMC11

`MiMC implementation using exponent of 11 and 51 rounds` -> `MiMC implementation using exponent of 11 and 51 rounds`

libzeth/circuits/mimc/mimc_mp.hpp
```diff
@@ -15,15 +15,16 @@ namespace libzeth
 /// MiMC_mp_gadget enforces correct computation of MiMC compression function
 /// based on a the Miyaguchi-Preneel compression construct and MiMC block cipher
 /// on Z_p with exponent 7 (and 91 rounds) p is given by the size(FieldT)
-template<typename FieldT> class MiMC_mp_gadget : public libsnark::gadget<FieldT>
+template<typename FieldT, typename PermutationT>
```
AntoineRondelet, line 18: Let's update the comment above

scripts/mimc_constraints.sage
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/sage
```
AntoineRondelet, line 1: What about moving this script in `libzeth/circuits/mimc`, next to https://github.com/clearmatics/zeth/blob/mimc-configurations/libzeth/circuits/mimc/mimc_round_constants_generation.py?

scripts/mimc_constraints.sage
```diff
@@ -0,0 +1,71 @@
...
+    # combine e_true_bits
+    return mults + mults_to_combine_num_powers(e_true_bits)
+
+
+def output_valid_config_and_constraints(r, log_2_r, e):
+    if 1 == gcd(r-1, e):
+        mults = compute_mults(e)
+        rounds = ceil(log_2_r / log(e, 2))
+        constraints = mults * rounds + 1
+        print(f"  e={e}, rounds={rounds}, mults={mults}, contraints={constraints}")
```
AntoineRondelet, line 47: I assume that your version of sage is >9.0, right?

zeth_contracts/contracts/MiMC31.sol
```diff
@@ -0,0 +1,77 @@
...
+// <https://eprint.iacr.org/2016/492.pdf>
+//
+// "One-way compression function"
+// Section: "Miyaguchi–Preneel"
+// <https://en.wikipedia.org/wiki/One-way_compression_function#Miyaguchi%E2%80%93Preneel>
+
+// Implementation of MiMC hash targetting BLS12-377 Fr. This means:
+//   r = 8444461749428370424248824938781546531375899335154063827935233455917409239041
+//   e (exponent) = 31, so that e=2^t-1 (t=5) and 1 == gcd(e, r-1)
+//   rounds = 73
```
AntoineRondelet, line 22: That should be 51

zeth_contracts/contracts/MiMC7.sol
```diff
@@ -4,77 +4,65 @@
...
-            // `outMiMCCipher = addmod(outPermutation, key, r)`.  And, the
-            // Myjaguchi-Prenell OWCF is ran: `addmod(addmod(outMiMCCipher,
-            // message, r), key, r)`.  Note that we have merged the key addition
-            // ( +key ) of the last round of MiMC with the Myjaguchi-Prenell
-            // step.
-            out := addmod(addmod(addmod(outPermutation, key, r), x, r), key, r)
+            // In MiMC, the final round output is summed with the round key
+            a := addmod(a, y, r)
+
+            // Myjaguchi-Prenell OWCF is then applied, which adds the key and
```
AntoineRondelet, line 63: Miyaguchi-Preneel

client/test_commands/test_mimc_contract.py
dtebbs: changed output string

client/tests/test_mimc.py
dtebbs: renamed variables

client/zeth/core/mimc.py
dtebbs: updated comment

client/zeth/core/mimc.py
dtebbs: fixed

scripts/mimc_constraints.sage
dtebbs: I'd prefer to keep these scripts in one place, rather than scattered amongst the source. It makes them less easy to find, and it's not clear to me that a mimc script belongs here rather than with the python or solidity implementation.  I realise it's not the same for the round constants script which is currently only used in the circuit implmentation (and in fact I hadn't noticed it there), but I'd rather move this into `/scripts` (or some subdirectory) unless you feel strongly.

Let me know.

scripts/mimc_constraints.sage
dtebbs: Yes.  I installed with `brew`, which only provides 9.2 at the current time.

zeth_contracts/contracts/MiMC31.sol
dtebbs: fixed. thanks.

zeth_contracts/contracts/MiMC7.sol
dtebbs: fixed typo.  thanks.

scripts/mimc_constraints.sage
dtebbs: (Added a commit to move the script to `scripts` so all scripts are there, but can easily be undone if you prefer to keep it where it was.)

scripts/mimc_constraints.sage
AntoineRondelet: Makes sense, I had an older sage installation (`<0.9`) which wasn't supporting python3 yet.

scripts/mimc_constraints.sage
AntoineRondelet: Sure, grouping them all in "scripts" is fine by me. That being said I think we would benefit from having more structure in the `scripts` folder. I added a `docker` subdirectory some time ago that aims to be the target for the docker entrypoints etc. I would do the same for the `test_` scripts: `cd scripts && mkdir tests && cp test_<NAME>.sh tests/<NAME>.sh`, and the same for `sage` scripts.
That's for another ticket.

AntoineRondelet: See #314

AntoineRondelet:
LGTM -  thanks @dtebbs

PR #318, opened 2020-11-24
Added some basic code style directives for solidity for consistency

AntoineRondelet:
After skimming through the solidity code, I realized that we had a lot of style inconsistencies throughout our files.
Here's an attempt to capture the general trend (but as above, several files don't fully comply with these style directives).

### DIFF
--- a/CODING_STANDARDS.md
+++ b/CODING_STANDARDS.md
@@ -13,6 +13,19 @@ New code should follow these rules, and old code should be changed over time to 
 
 We adhere to the `pep8` standard, using `flake8`, `pylint` and `mypy` to catch further programming errors or standard violations. Type hints must be used everywhere.
 
+## Solidity
+
+The Solidity code follows the following conventions:
+
+- Contracts are named following `PascalCase`. Contract names and contract files have the same name (e.g. the contract `MyContract` is implemented in `MyContract.sol`). Same applies for libraries.
+- Testing contract names end with `_test`.
+- Documentation code comments follow [NatSpec](https://docs.soliditylang.org/en/v0.7.1/natspec-format.html#natspec).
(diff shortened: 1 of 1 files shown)


CODING_STANDARDS.md
```diff
@@ -13,6 +13,19 @@ New code should follow these rules, and old code should be changed over time to
...
+## Solidity
+
+The Solidity code follows the following conventions:
+
+- Contracts are named following `PascalCase`. Contract names and contract files have the same name (e.g. the contract `MyContract` is implemented in `MyContract.sol`). Same applies for libraries.
+- Testing contract names end with `_test`.
+- Documentation code comments follow [NatSpec](https://docs.soliditylang.org/en/v0.7.1/natspec-format.html#natspec).
+- Variable names and argument names are in `snake_case` (e.g. `my_variable`) except *state variables* which are prefixed with `_` (e.g. `_my_state_var`).
+- Constant names are in `UPPER_SNAKE_CASE`.
+- Function names are in `snake_case`.
```
dtebbs, line 25: `Method names`, maybe?

dtebbs (commented):
One optional comment, but LGTM.

dtebbs (approved):
(Resubmitting - I meant to click "approve" in the first review)

CODING_STANDARDS.md
AntoineRondelet: I don't have strong preferences (methods tend to denote "object's functions" in OOP, but a method truly is a function). However, they seem to stick to "function" in solidity (see e.g. here: https://docs.soliditylang.org/en/v0.7.1/contracts.html?highlight=method#functions) rather than "methods" for contracts.

dtebbs: For some reason, I thought functions *had* to be on contracts, in which case "methods" would seem more suitable. But I now see from the specs that functions can actually be standalone, so "functions" is more natural.

Issue #304, opened 2020-10-28
Extend CI to build project with sanitizers and set up code scaning


AntoineRondelet:
As described in the title, it would be nice to support compilation with sanitizers to detect potential unsound memory accesses and race conditions on the CI server. Likewise, Github seems to support "code scanning" out of the box to detect potential  vulnerabilities in the code.
Extending the CI to automate these type of checks as much as possible is a good way to maintain high security standards (along with high "code quality"/"stylistic"/"consistency" standards as currently ensured by the use of `clang-format`/`clang-format`)

AntoineRondelet:
See also: https://github.com/snyk/actions

PR #319, opened 2020-11-24
Added minimal configuration for code scanning

AntoineRondelet:
See: https://github.com/clearmatics/zeth/issues/304


.github/workflows/code-scanning.yml
```diff
@@ -0,0 +1,64 @@
+name: "code-scanning"
+
+on:
+  push:
+  pull_request:
```
AntoineRondelet, line 5: This workflow is taking quite some time to complete. It makes more sense to switch to `on push/on pull_request` for the `master` branch only.

dtebbs: Or maybe `develop` and `on: push` only, so we can catch issues before anything gets into `master`?

.github/workflows/code-scanning.yml
```diff
@@ -0,0 +1,64 @@
...
+  # Schedule code scanning
```
dtebbs, line 6: To be removed?

.github/workflows/code-scanning.yml
```diff
@@ -0,0 +1,64 @@
...
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: Install Dependencies
+        run: brew install ${MACOS_BREW_PACKAGES}
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v1
+        with:
+          languages: cpp
+        # Autobuild for compiled languages.
```
dtebbs, line 29: remove?

.github/workflows/code-scanning.yml
dtebbs: The mac build seems to be very slow in this branch (which seems to be the reason for this workflow taking so long).

.github/workflows/code-scanning.yml
AntoineRondelet: Hum I actually want to keep this comment around. May be useful to set a cron job to scan the code at given intervals in addition to scan in `push/pull-request` (and I am sure I will forget that we can schedule cron jobs in workflows if I remove this comment)

.github/workflows/code-scanning.yml
AntoineRondelet: > Or maybe develop and on: push only, so we can catch issues before anything gets into master?

Configuring the `on: pull-request` for `master` will also catch any issue before we merge to master, but yes, we can also add `on:push` for `develop`. The main point is to keep an sufficiently performant CI (this workflow is taking ages unfortunately)

.github/workflows/code-scanning.yml
AntoineRondelet: Ah yes, this one needs to go away, thanks I forgot to remove it.

AntoineRondelet:
Related to: https://github.com/github/docs/pull/1637

AntoineRondelet:
Seems like the CI is not triggered (probably because of https://www.githubstatus.com/incidents/sxthxvdlljk4). Let's investigate why the macOS build took so long for the last CI run (https://github.com/clearmatics/zeth/pull/319#discussion_r533242841). This will educate when we set up code scanning (https://github.com/clearmatics/zeth/pull/319#discussion_r531716108)

.github/workflows/code-scanning.yml
dtebbs: This may be better now.  I hit the same thing (mac build taking ~1hour).  Seemed to be caused by the initial `brew install` command, but the issue magically resolved itself when I tried to add logging to investigate further.

AntoineRondelet: Seems like the macOS build is faster now indeed. The scanning job still takes longer (~30mins) but this is much better than the 1h+ running time we had a few days back.

AntoineRondelet: I adjusted the scope of the code-scanning workflow to only run `on: push` on `develop` and `on: pull-request` on `master`. Does that sound like a right policy to you @dtebbs ? (the scanning jobs are still ~3x longer to run than the other ones so may be worth running them less frequently)

dtebbs: Yes, that sounds fine.  It would be nice to have it run *before* we merge to develop, but as you say the longer run time makes that difficult. This seems like a good compromise.

dtebbs (approved):
LGTM

Issue #331, opened 2020-12-17
(Temporary) CI build failure in libsodium


dtebbs:
All builds are failing:
```
configure: error: cannot run /bin/sh build-aux/config.sub
```

This file seems to be downloaded (even though it is actually in the repo), and contains:
```
<html>
<head><title>502 Bad Gateway</title></head>
<body bgcolor="white">
<center><h1>502 Bad Gateway</h1></center>
<hr><center>nginx/1.4.6 (Trisquel GNU/Linux)</center>
</body>
</html>
```

AntoineRondelet:
Closing this issue since the corresponding PR has been merged.

PR #332, opened 2020-12-17
Fix intermittent CI failure

dtebbs:
See #331


AntoineRondelet:
Thanks @dtebbs I am seeing this one a bit late.. but it now seems to be working well (the CI works on https://github.com/clearmatics/zeth/pull/333).
The error you mentioned in #331 was a server-side error. Seems like it's been fixed now.

dtebbs:
Could still be worth not downloading during the build.  This has happened a few times in the past as well.

AntoineRondelet:
> Could still be worth not downloading during the build. This has happened a few times in the past as well.

Alright.

Issue #321, opened 2020-11-25
Bring the solidity code up to date with the coding standards


AntoineRondelet:
The PR https://github.com/clearmatics/zeth/pull/318 added some basic and missing code standards for the solidity code. Several files are not compliant with it (in fact, as of now some contracts code styles conflict) and follow different conventions. Let's bring all these files up to date with the solidity code standards to have a consistent code.

In the same time, we may want to use solhint (https://github.com/protofire/solhint) as solidity linter to enforce best practices (code/style) and check these on the CI server.

AntoineRondelet:
Despite #333 being merged, this ticket remains opened since the smart contract code is not fully compliant with the solidity standards (doing so would have cause too many changes for a single PR). This will be done incrementally. As such, this ticket remains until the whole code satisfies the solidity coding standards.

AntoineRondelet:
https://github.com/clearmatics/zeth/pull/361 has brought the contract code up to date with the solidity standards overall. We made amendments to the coding standards of solidity to fit our needs, but the functions' interfaces etc are now compliant.
Closing this issue as a consequence.

Issue #192, opened 2020-04-08
Update documentation to follow NatSpec guidelines


rrtoledo:
Let's follow [NatSpec](https://solidity.readthedocs.io/en/v0.6.4/natspec-format.html) for "documentation comments"

AntoineRondelet:
Thread: https://github.com/clearmatics/zeth/pull/157#discussion_r402803686

AntoineRondelet:
Contained in https://github.com/clearmatics/zeth/issues/321

AntoineRondelet:
The existing smart-contracts have been updated to follow NatSpec. Closing this issue as a consequence.

PR #333, opened 2020-12-17
Shallow clean-up of the solidity code

AntoineRondelet:
Related to: https://github.com/clearmatics/zeth/issues/321 and https://github.com/clearmatics/zeth/issues/192

This PR:
- Adjusts the Zeth solidity coding standards to rather align with the "language standards" proposed by the community and the solidity team. The driving factors for that are:
    - If there is a standard we should do our best to follow it (good practice) - it removes friction as to "what looks better" and just follows the norm.
    - Removes sources of inconsistencies between the code we write and the code we import (which catches the eye pretty badly when we call a function from an imported contract, like an ERC token for instance)
- Renames all solidity constants in `UPPER_SNAKE_CASE`
- Adds explicit scopes to all functions and storage variables (this should become systematic from now on as more verbosity with regard to the scopes can only be better)
- Switches the linter to `solhint` (`solium` then replaced by `ethlint` seems to be poorly maintained, last release was more than a year ago. `solhint` is actively maintained offers a good configuration set and is used by openZeppelin (https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/.solhint.json) and other serious projects). Several checks are voluntarily disabled for now because I didn't want to end up with a PR that shuffles the entire code base (renaming public functions, contract names etc => cascading these changes to the python code etc which will become a nightmare to review. Let's take it incrementally to avoid adding mistakes to the code base and breaking things)
- Uses `NatSpec` for documentation comments
- Orders functions (as per: https://docs.soliditylang.org/en/develop/style-guide.html#order-of-functions)
- Fixes the order of the modifiers

TODO:
- [x] Fix the `test_merkle_tree_contract.py` test (we have a `MerkleTreeMiMC7_test` test contract but its import is broken since `MerkleTreeMiMC7.sol` does not exist anymore)
- [x] Fix the `test_contract_base_mixer.py` test. It is now broken because the scope of some functions (e.g. `test_assemble_nullifiers`) has been changed in this PR, so these functions are not in the ABI anymore. Since we should not expose unnecessary function simply to be able to call them in the test scripts, let's add a `MixerBase_test.sol` "test proxy contract" to re-enable the tests despite the functions visibility change. - Fixed in #338


zeth_contracts/contracts/BLS12_377_test.sol
```diff
@@ -9,39 +9,39 @@ contract BLS12_377_test
...
+    uint256 internal constant SCALAR_BYTES = SCALAR_WORDS * 32; // 32 (0x20)
 
-    uint256 constant g1CoordWords = 2;
-    uint256 constant g1CoordBytes = g1CoordWords * 32; // 64 (0x40)
-    uint256 constant g1PointWords = 2 * g1CoordWords;
-    uint256 constant g1PointBytes = g1PointWords * 32; // 128 (0x80)
+    uint256 internal constant G1_COORD_WORDS = 2;
+    uint256 internal constant G1_COORD_BYTES = G1_COORD_WORDS * 32; // 64 (0x40)
+    uint256 internal constant G1_POINT_WORDS = 2 * G1_COORD_WORDS;
+    uint256 internal constant G1_POINT_BYTES = G1_POINT_WORDS * 32; // 128 (0x80)
```
AntoineRondelet, line 18: Unused?

zeth_contracts/contracts/BLS12_377_test.sol
```diff
@@ -9,39 +9,39 @@ contract BLS12_377_test
...
-    uint256 constant scalarBytes = scalarWords * 32; // 32 (0x20)
+    uint256 internal constant SCALAR_WORDS = 1;
+    uint256 internal constant SCALAR_BYTES = SCALAR_WORDS * 32; // 32 (0x20)
 
-    uint256 constant g1CoordWords = 2;
-    uint256 constant g1CoordBytes = g1CoordWords * 32; // 64 (0x40)
-    uint256 constant g1PointWords = 2 * g1CoordWords;
-    uint256 constant g1PointBytes = g1PointWords * 32; // 128 (0x80)
+    uint256 internal constant G1_COORD_WORDS = 2;
+    uint256 internal constant G1_COORD_BYTES = G1_COORD_WORDS * 32; // 64 (0x40)
```
AntoineRondelet, line 16: Unused?

zeth_contracts/contracts/BLS12_377_test.sol
```diff
@@ -9,39 +9,39 @@ contract BLS12_377_test
     // In many cases, these numbers must be used as literals in the assembly
     // code.
 
-    uint256 constant scalarWords = 1;
-    uint256 constant scalarBytes = scalarWords * 32; // 32 (0x20)
+    uint256 internal constant SCALAR_WORDS = 1;
+    uint256 internal constant SCALAR_BYTES = SCALAR_WORDS * 32; // 32 (0x20)
```
AntoineRondelet, line 13: Unused?

zeth_contracts/contracts/BLS12_377_test.sol
```diff
@@ -9,39 +9,39 @@ contract BLS12_377_test
...
+    uint256 internal constant G1_POINT_BYTES = G1_POINT_WORDS * 32; // 128 (0x80)
 
-    uint256 constant g2CoordWords = 4;
-    uint256 constant g2CoordBytes = g2CoordWords * 32; // 128 (0x80)
-    uint256 constant g2PointWords = 2 * g2CoordWords;
-    uint256 constant g2PointBytes = g2PointWords * 32; // 256 (0x100)
+    uint256 internal constant G2_COORD_WORDS = 4;
+    uint256 internal constant G2_COORD_BYTES = G2_COORD_WORDS * 32; // 128 (0x80)
+    uint256 internal constant G2_POINT_WORDS = 2 * G2_COORD_WORDS;
+    uint256 internal constant G2_POINT_BYTES = G2_POINT_WORDS * 32; // 256 (0x100)
```
AntoineRondelet, line 23: Unused?

zeth_contracts/contracts/BLS12_377_test.sol
```diff
@@ -9,39 +9,39 @@ contract BLS12_377_test
...
+    uint256 internal constant G1_COORD_BYTES = G1_COORD_WORDS * 32; // 64 (0x40)
+    uint256 internal constant G1_POINT_WORDS = 2 * G1_COORD_WORDS;
+    uint256 internal constant G1_POINT_BYTES = G1_POINT_WORDS * 32; // 128 (0x80)
 
-    uint256 constant g2CoordWords = 4;
-    uint256 constant g2CoordBytes = g2CoordWords * 32; // 128 (0x80)
-    uint256 constant g2PointWords = 2 * g2CoordWords;
-    uint256 constant g2PointBytes = g2PointWords * 32; // 256 (0x100)
+    uint256 internal constant G2_COORD_WORDS = 4;
+    uint256 internal constant G2_COORD_BYTES = G2_COORD_WORDS * 32; // 128 (0x80)
```
AntoineRondelet, line 21: Unused?

zeth_contracts/contracts/BW6_761_test.sol
```diff
@@ -9,18 +9,18 @@ contract BW6_761_test
...
     // code.
 
-    uint256 constant scalarWords = 2;
-    uint256 constant scalarBytes = scalarWords * 32; // 64 (0x40)
-    uint256 constant coordWords = 3;
-    uint256 constant coordBytes = 3 * 32; // 96 (0x60)
-    uint256 constant pointWords = 2 * coordWords;
-    uint256 constant pointBytes = pointWords * 32; // 192 (0xc0)
+    uint256 internal constant SCALAR_WORDS = 2;
+    uint256 internal constant SCALAR_BYTES = SCALAR_WORDS * 32; // 64 (0x40)
```
AntoineRondelet, line 13: Unused?

zeth_contracts/contracts/BW6_761_test.sol
```diff
@@ -9,18 +9,18 @@ contract BW6_761_test
...
+    uint256 internal constant COORD_WORDS = 3;
+    uint256 internal constant COORD_BYTES = 3 * 32; // 96 (0x60)
```
AntoineRondelet, line 15: Unused?

zeth_contracts/contracts/BW6_761_test.sol
```diff
@@ -9,18 +9,18 @@ contract BW6_761_test
...
+    uint256 internal constant POINT_WORDS = 2 * COORD_WORDS;
+    uint256 internal constant POINT_BYTES = POINT_WORDS * 32; // 192 (0xc0)
```
AntoineRondelet, line 17: Unused?

AntoineRondelet:
@dtebbs there are some solidity variables that are seemingly unused. Could you double check this before we go further and consider removing them? (Are you using them somewhere I missed?)

zeth_contracts/contracts/BW6_761_test.sol
AntoineRondelet: Btw all these can be made `private` I think. Would be good to tighten the scopes.

dtebbs:
> @dtebbs there are some solidity variables that are seemingly unused. Could you double check this before we go further and consider removing them? (Are you using them somewhere I missed?)

@AntoineRondelet  Yes, they do seem to unused.  Good catch.

zeth_contracts/.solhint.json
```diff
@@ -0,0 +1,16 @@
...
+        "func-name-mixedcase": "off",
+        "contract-name-camelcase": "off",
+        "no-empty-blocks": "off",
+        "compiler-version": "off",
+        "reason-string": "off",
+        "avoid-suicide": "error",
+        "no-inline-assembly": "off",
+        "avoid-sha3": "warn",
+        "func-visibility": "error"
+    }
```
dtebbs, line 15: Just FYI, I tried adding `"max-line-length": ["error",80]`, but this catches comments (e.g. long URLs).  Would be nice to have that at some point, though, perhaps using `// solhint-diable-next-line ...` where long lines are unavoidable.

zeth_contracts/package.json
```diff
@@ -10,16 +10,16 @@
     "deploy": "truffle deploy",
     "test": "truffle test",
     "debug": "truffle debug",
-    "check": "node_modules/solium/bin/solium.js --dir contracts"
+    "check": "node_modules/solhint/solhint.js 'contracts/**/*.sol'"
```
dtebbs, line 13: This command does not exit with error for warnings (I tried changing a constant to lower case).  This means that the CI will never detect these issues.

dtebbs (commented):
Looks good.  Would be nice to have the CI catch violations.  (I'll look into that)

dtebbs:
I've tried to address my comments in #334

zeth_contracts/.solhint.json
AntoineRondelet: Thanks #334 was merged in this branch.

AntoineRondelet:
Note: the `test_contract_base_mixer.py` test is broken. To properly fix it we would need to use a "test proxy contract", but this is for later (see #335). For now, let's just roll back and set the scopes of the tested functions back to `public` and update the tests to make sure that the tested functions behave correctly.

dtebbs:
(Worth rebasing this onto origin/dvelop before merging)

dtebbs:
#338 fixes the tests and provides public wrappers around the internal functions (so no need to adjust this code)

PR #334, opened 2020-12-18
Solidity standards dt

dtebbs:
A few additions to the solidity changes
- [X] detect violations
- [X] enable line-length checks
- [X] remove remaining solium references

### DIFF
--- a/zeth_contracts/contracts/Pairing.sol
+++ b/zeth_contracts/contracts/Pairing.sol
@@ -11,6 +11,7 @@ pragma solidity ^0.5.0;
 /// implementations:
 ///   https://github.com/ethereum/go-ethereum/blob/master/core/vm/contracts.go gas
 /// and costs:
+// solhint-disable-next-line
 ///   https://github.com/ethereum/go-ethereum/blob/master/params/protocol_params.go
 library Pairing {
 
@@ -35,20 +36,20 @@ library Pairing {
     // Return the generator of G2
     function P2() internal pure returns (G2Point memory) {
         return G2Point(
-            // solium-disable-next-line
+            // solhint-disable-next-line
             11559732032986387107991004021392285783925812861821192530917403151452391805634,
-            // solium-disable-next-line
+            // solhint-disable-next-line
             10857046999023057135944570762232829481370756359578518086990519993285655852781,
-            // solium-disable-next-line
+            // solhint-disable-next-line
             4082367875863433681332203403145435568316851327593401208105741076214120093531,
-            // solium-disable-next-line
+            // solhint-disable-next-line
             8495653923123431417604973247489272438418190587263600148770280649306958101930);
     }
 
     // Return the negation of p, i.e. p.add(p.negate()) should be zero.
     function negate(G1Point memory p) internal pure returns (G1Point memory) {
         // The prime q in the base field F_q for G1
-        // solium-disable-next-line
+        // solhint-disable-next-line
         uint256 q = 21888242871839275222246405745257275088696311157297823662689037894645226208583;
         if (p.X == 0 && p.Y == 0)
             return G1Point(0, 0);
@@ -74,7 +75,8 @@ library Pairing {
         }
         require(
             success,
-            "Call to the bn256Add precompiled failed (probably an out of gas error?)"
+            "Call to the bn256Add precompiled failed (probably an out of gas "
+            "error?)"
         );
     }
 
@@ -130,6 +132,7 @@ library Pairing {
         bool success;
         assembly {
             // bn256Pairing precompiled:
+            // solhint-disable-next-line
             //   https://github.com/ethereum/go-ethereum/blob/v1.7.0/core/vm/contracts.go#L59
             //
             // The bn256Pairing precompiled takes an input of size N * 192 (a
@@ -149,13 +152,21 @@ library Pairing {
             // any number of pairs (g1, g2) \in G1 x G2.  To check something in
             // the form: e(g1, g2) = e(g'1, g'2), we need to call the
             // precompiled bn256Pairing on input [(g1, g2), (neg(g'1), g'2)]
-            success := call(sub(gas, 2000), 8, 0, add(input, 0x20), mul(inputSize, 0x20), out, 0x20)
+            success := call(
+              sub(gas, 2000),
+              8,
+              0,
+              add(input, 0x20),
+              mul(inputSize, 0x20),
+              out,
+              0x20)
             // Use "invalid" to make gas estimation work
             //switch success case 0 { invalid }
         }
         require(
             success,
-            "Call to the bn256Pairing precompiled failed (probably an out of gas error?)"
+            "Call to the bn256Pairing precompiled failed (probably an out of "
+            "gas error?)"
         );
 
         return out[0] != 0;
--- a/zeth_contracts/contracts/Pairing.sol
+++ b/zeth_contracts/contracts/Pairing.sol
@@ -6,11 +6,11 @@ pragma solidity ^0.5.0;
 
 /// Several pairing-related utility functions.
 ///
-/// Precompiled contract details (bn256Add, bn256ScalarMul, bn256Pairing) can be
-/// found at the following links:
+/// Precompiled contract details (bn256Add, bn256ScalarMul, bn256Pairing) can
+/// be found at the following links:
 /// implementations:
-///   https://github.com/ethereum/go-ethereum/blob/master/core/vm/contracts.go gas
-/// and costs:
+///   https://github.com/ethereum/go-ethereum/blob/master/core/vm/contracts.go
+/// gas and costs:
 // solhint-disable-next-line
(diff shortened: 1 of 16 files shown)


dtebbs:
@AntoineRondelet we might want to consider removing the line-length change.  Let me know what you think.

zeth_contracts/package.json
```diff
@@ -10,7 +10,7 @@
     "deploy": "truffle deploy",
     "test": "truffle test",
     "debug": "truffle debug",
-    "check": "node_modules/solhint/solhint.js 'contracts/**/*.sol'"
+    "check": "node_modules/solhint/solhint.js --max-warnings 0 'contracts/**/*.sol'"
```
AntoineRondelet, line 13: Thanks @dtebbs good trick to make this fail on the CI

AntoineRondelet (commented):
Thanks for adding the `solhint-disable` comments where needed (and removing the reference to `solium` in the code).
Overall I think the `line-length` check alters readability quite a lot TBH. What do you think?

As discussed off-line, the tricky thing with following PEP8-like max line length (https://www.python.org/dev/peps/pep-0008/#maximum-line-length) as initially suggested by the solidity standard, is that we end up with a very vertical code (function signatures can be very very lengthy in solidity since they contain function names, visibilty (internal/public etc.), view/pure, modifiers, returned values etc.).
Some solidity projects like OpenZeppelin (see e.g. their ERC code https://github.com/OpenZeppelin/openzeppelin-contracts/tree/master/contracts/token/ERC20) have disabled line length checks to - I believe - tradeoff with readability (others like Gnosis have set the line length to a less conservative chars nb https://github.com/gnosis/dex-contracts/blob/master/.solhint.json#L11-L14). On our end I think there is value to try to avoid overly lengthy lines, but being very strict with the `<80` chars policy seems to be altering readability quite a lot here. I'd suggest to relax this for now.

zeth_contracts/contracts/Tokens.sol
```diff
@@ -25,16 +27,16 @@ contract ERC223ReceivingContract {
         bytes4 sig;
     }
 
-    function tokenFallback(address from, uint256 value, bytes memory data)
-        public
-        pure
+    function
+    tokenFallback(address from, uint256 value, bytes memory data) public pure
```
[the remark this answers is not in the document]
dtebbs: Oops.  That was an accident.  Fixed now.

dtebbs:
(I tend to think that if the code looks vertical it's because there is a lot happening, and in most cases it would look more confusing as fewer lines.)

> but it this truly alters readability I'd be inclined to relax the line length.

So, are you suggesting a longer line length, or just revert the whole thing?

TBH, I don't see too much of a readability issue yet in this code. To me the main issue was the pragmas for comments (hopefully this will be fixed in a future version of solhint), but on balance I'd prefer to avoid ambiguity here (i.e. having some code staying within 80 and some using longer lines - if I see a long line in a review it becomes a matter of opinion whether it should be broken or not).

Having said that, we could equally just revert for now and deal with it another time.  Perhaps that's the best approach.
I've reordered the commits so the line-length one can be easily excluded.

AntoineRondelet:
> (I tend to think that if the code looks vertical it's because there is a lot happening, and in most cases it would look more confusing as fewer lines.)
> 
> > but it this truly alters readability I'd be inclined to relax the line length.
> 
> So, are you suggesting a longer line length, or just revert the whole thing?
> 
> TBH, I don't see too much of a readability issue yet in this code. To me the main issue was the pragmas for comments (hopefully this will be fixed in a future version of solhint), but on balance I'd prefer to avoid ambiguity here (i.e. having some code staying within 80 and some using longer lines - if I see a long line in a review it becomes a matter of opinion whether it should be broken or not).
> 
> Having said that, we could equally just revert for now and deal with it another time. Perhaps that's the best approach.
> I've reordered the commits so the line-length one can be easily excluded.

If I strictly follow the solidity standards things like [this](https://github.com/clearmatics/zeth/blob/solidity-standards-dt/zeth_contracts/contracts/Pghr13AltBN128Mixer.sol#L121-L125):
```javascript
        if (!Pairing.pairingProd3(
                proof.K, vk.gamma,
                Pairing.negate(Pairing.add(vk_x, Pairing.add(proof.A, proof.C))),
                vk.gammaBeta2,
                Pairing.negate(vk.gammaBeta1), proof.B)
        ) {
            return 4;
        }
```
would need to be broken down to add linebreaks between function arguments and polish the indent. Furthermore, the if condition is already quite long which I personally (this is subjective) am not necessarily keen on. 

That's not a major deal. Let's keep in that way if you think it's fine. I don't have strong opinions w.r.t such questions around the length of the lines. If we can remove dev subjectivity by adopting a "standard" that's fine. Let's follow the solidity standards then and add the "<80" line length directive to the linter config.

zeth_contracts/contracts/Pairing.sol
```diff
@@ -149,13 +152,21 @@ library Pairing {
             // any number of pairs (g1, g2) \in G1 x G2.  To check something in
             // the form: e(g1, g2) = e(g'1, g'2), we need to call the
             // precompiled bn256Pairing on input [(g1, g2), (neg(g'1), g'2)]
-            success := call(sub(gas, 2000), 8, 0, add(input, 0x20), mul(inputSize, 0x20), out, 0x20)
+            success := call(
+              sub(gas, 2000),
```
AntoineRondelet, line 156: Seems like you used 2 spaces for the indent, should be 4

zeth_contracts/contracts/Pairing.sol
```diff
@@ -149,13 +152,21 @@ library Pairing {
...
+              8,
+              0,
+              add(input, 0x20),
+              mul(inputSize, 0x20),
+              out,
+              0x20)
```
AntoineRondelet, line 162: ` 0x20)` -> ` 0x20\linebreak\indent)`

AntoineRondelet:
Few things though:
- Let's avoid things like:
```javascript
function test(
    arg1,
    arg2,
    ...,
    argN);
```

Seems like it should rather be (following the standard):
```javascript
function test(
    arg1,
    arg2,
    ...,
    argN
);
```
(same for other things like `require`, `events` etc.)

- We should be avoiding things like
```javascript
function test(
    arg1, arg2,
    ..., argN
);
```
And rather break after all args in this setting

dtebbs:
> > (I tend to think that if the code looks vertical it's because there is a lot happening, and in most cases it would look more confusing as fewer lines.)
> > > but it this truly alters readability I'd be inclined to relax the line length.
> > 
> > 
> > So, are you suggesting a longer line length, or just revert the whole thing?
> > TBH, I don't see too much of a readability issue yet in this code. To me the main issue was the pragmas for comments (hopefully this will be fixed in a future version of solhint), but on balance I'd prefer to avoid ambiguity here (i.e. having some code staying within 80 and some using longer lines - if I see a long line in a review it becomes a matter of opinion whether it should be broken or not).
> > Having said that, we could equally just revert for now and deal with it another time. Perhaps that's the best approach.
> > I've reordered the commits so the line-length one can be easily excluded.
> 
> If I strictly follow the solidity standards things like [this](https://github.com/clearmatics/zeth/blob/solidity-standards-dt/zeth_contracts/contracts/Pghr13AltBN128Mixer.sol#L121-L125):
> 
> ```js
>         if (!Pairing.pairingProd3(
>                 proof.K, vk.gamma,
>                 Pairing.negate(Pairing.add(vk_x, Pairing.add(proof.A, proof.C))),
>                 vk.gammaBeta2,
>                 Pairing.negate(vk.gammaBeta1), proof.B)
>         ) {
>             return 4;
>         }
> ```
> 
> would need to be broken down to add linebreaks between function arguments and polish the indent. Furthermore, the if condition is already quite long which I personally (this is subjective) am not necessarily keen on.
> 

Right.  That's a good example where some extra local variables could really aid readability.  I also had to look pretty closely to notice that some arguements are sharing lines.

> That's not a major deal. Let's keep in that way if you think it's fine. I don't have strong opinions w.r.t such questions around the length of the lines. If we can remove dev subjectivity by adopting a "standard" that's fine. Let's follow the solidity standards then and add the "<80" line length directive to the linter config.

OK.  I'll switch it to <80 then.

dtebbs:
Changes for line-length <80 should now be pushed.

zeth_contracts/contracts/Pairing.sol
dtebbs: Added extra indent.

zeth_contracts/contracts/Pairing.sol
dtebbs: Moved parenthesis to the next line.

zeth_contracts/contracts/Pairing.sol
AntoineRondelet: Thanks

AntoineRondelet (approved):
LGTM

Issue #335, opened 2020-12-18
Tigthen function visibilities/scopes in solidity contracts


AntoineRondelet:
As of now, many functions are kept `public` on smart contracts to add them to the ABIs and allow the testing scripts to call them (for instance, see the `assemble_nullifier`, `assemble_hsig` etc function on the `MixerBase` contract).
The ABI should only contain functions exposed to the end user. As such, let's switch the scope of `public` functions to become (at least) `internal`, and use "test contracts" to wrap these internal functions with public ones. E.g.

```javascript
contract MyContract{

function myFunctionToTest() internal {...}

}
```

and use the following wrapping contract to execute the tests:
```javascript
contract MyContractTest is MyContract {

function myTestFunction() public {
    return myFunctionToTest();
}

}
```

Then, we can call `MyContractTest` from the python testing scripts.

**Note:** We may also want to move more of that test logic to solidity and expose a single entry point for all tests in the "test contract".

AntoineRondelet:
(Also related to https://github.com/clearmatics/zeth/pull/141)

AntoineRondelet:
Tackled as part of #361

PR #338, opened 2020-12-18
Contract tests fix (depends on #333)

dtebbs:
Fix and clean up some contract tests (#333)

- [x] Fix values and fully test residual bits
- [x] clean up variables
- [x] use a wrapping contract and rename to the actual contract being tested
- [x] remove dependency on prover server

(Does not move test logic into contract, but that may involve more complexity so maybe better addressed in the future)
Related to #335


zeth_contracts/contracts/MerkleTreeMiMC7_test.sol
```diff
@@ -4,7 +4,8 @@
 
 pragma solidity ^0.5.0;
 
-import "./MerkleTreeMiMC7.sol";
+import "./BaseMerkleTree.sol";
```
AntoineRondelet, line 7: This is already fixed on #333

client/test_commands/test_altbn128_mixer_base.py
```diff
@@ -0,0 +1,148 @@
...
+# SPDX-License-Identifier: LGPL-3.0+
+
+from zeth.core.constants import \
+    JS_INPUTS, ZETH_PUBLIC_UNIT_VALUE, ZETH_MERKLE_TREE_DEPTH
+import test_commands.mock as mock
+from typing import Any
+
+# pylint: disable=line-too-long
+
+# TODO: These test is specific to AltBN128MixerBase, however the mixer that is
```
AntoineRondelet, line 14: `These test is` -> `These tests are`

client/test_commands/test_altbn128_mixer_base.py
```diff
@@ -0,0 +1,148 @@
...
+    "0000000011101110111011101110111011101110111011101110111011101110",  # vout
+    2)
+
+PACKED_PRIMARY_INPUTS = \
+    [ROOT] + COMMITMENTS + NULLIFIERS + [HSIG] + HTAGS + [RESIDUAL_BITS]
+
+
+def test_assemble_nullifiers(mixer_instance: Any) -> None:
+    # Test retrieving nullifiers
+    print("--- testing ", "test_assemble_nullifiers")
```
AntoineRondelet, line 97: `test_assemble_nullifiers` -> `assemble_nullifier_test` (same for the functions below)

client/test_commands/test_altbn128_mixer_base.py
```diff
@@ -0,0 +1,148 @@
...
+    print("--- testing ", "test_assemble_hsig")
+    res = mixer_instance.functions.\
+        assemble_hsig_test(PACKED_PRIMARY_INPUTS).call()
+    hsig = int.from_bytes(res, byteorder="big")
+    assert hsig == HSIG, f"expected: {HSIG}, got {hsig}"
+
+
+def test_assemble_vpub(mixer_instance: Any) -> None:
+    # Test retrieving commitments
+    print("--- testing ", "test_assemble_vpub")
```
AntoineRondelet, line 116: `test_..._vpub` -> `assemble_public_values_test`

client/test_commands/test_altbn128_mixer_base.py
```diff
@@ -0,0 +1,148 @@
...
+    print("Deploying AltBN128MixerBase_test.sol")
+    _web3, eth = mock.open_test_web3()
+    deployer_eth_address = eth.accounts[0]
+    _mixer_interface, mixer_instance = mock.deploy_contract(
+        eth,
+        deployer_eth_address,
+        "AltBN128MixerBase_test",
+        {
+            'mk_depth': ZETH_MERKLE_TREE_DEPTH,
+        })
```
AntoineRondelet, line 135: What about having a linebreak before closing the function call here?

zeth_contracts/contracts/AltBN128MixerBase_test.sol
```diff
@@ -0,0 +1,57 @@
...
+contract AltBN128MixerBase_test is AltBN128MixerBase
+{
+    constructor(uint256 mk_depth)
+        public
+        AltBN128MixerBase(mk_depth, address(0), new uint256[](0))
+    {
+    }
+
+    function assemble_public_values_test(uint256 residual_bits)
+        public pure
```
AntoineRondelet, line 22: I think `pure` should be on the next line

client/test_commands/test_altbn128_mixer_base.py
dtebbs: Should this not be the name of the test? `assemble_nullifier_test` is the method name I happen to have used to wrap the internal methods.  Those solidity method names are more of an implementation detail (I was actually about to try reusing the same names - not sure if that works in solidity), whereas I guess the original `test_assemble_nullifiers` refers to the name of the test (the python function name).

dtebbs: I'll switch  it to: `print("--- test_assemble_nullifiers")`, which is hopefully a bit clearer.

client/test_commands/test_altbn128_mixer_base.py
dtebbs: We could switch to that style, but I think most existing code keeps the closign parenthesis on the same line.

client/test_commands/test_altbn128_mixer_base.py
dtebbs: fixed.  thanks.

client/test_commands/test_altbn128_mixer_base.py
dtebbs: (changed to clearer print statements)

zeth_contracts/contracts/AltBN128MixerBase_test.sol
dtebbs: moved to next line

client/test_commands/test_altbn128_mixer_base.py
dtebbs: (as above.  `test_assemble_vpub` seems more suitable for printing.)

client/test_commands/test_altbn128_mixer_base.py
AntoineRondelet: > I'll switch it to: print("--- test_assemble_nullifiers"), which is hopefully a bit clearer.

Thanks, that's a bit clearer (the `testing` prefix suggested that we ran a function named `test_assemble_nullifiers` which was not the case.).

> (changed to clearer print statements. Just FYI, I tried to keep the same names, but solidity does not seem to support overloading in this way, so I've kept the actual entry points as ...._test in the wrapper contract.)

Yes, let's keep the function names as they are If you changed the logs as per your comment above, that's fine. Thanks.

client/test_commands/test_altbn128_mixer_base.py
AntoineRondelet: Ah right, let's stay consistent with existing code then.

zeth_contracts/contracts/MerkleTreeMiMC7_test.sol
dtebbs: I've rebased onto the target branch.  thanks.

zeth_contracts/contracts/AltBN128MixerBase_test.sol
```diff
@@ -0,0 +1,60 @@
...
+        pure
+        returns(bytes32 nf)
+    {
+        return assemble_nullifier(index, primary_inputs);
+    }
+
+    // Dummy implementation of abstract function
+    function verify_zk_proof(
+        uint256[] memory proof,
+        uint256[NUM_INPUTS] memory inputs
```
AntoineRondelet, line 53: `proofs` and `inputs` are not used which makes the linter complain and causes the CI to fail. Let's either use them in a dummy condition, or switch the linter off for this block.

dtebbs: Using the `/* inputs */` approach, similar to C++ seems to satisfy the linter, if that works for you.

AntoineRondelet: Oh great, yes that's better.

AntoineRondelet (approved):
Thanks @dtebbs LGTM

PR #341, opened 2020-12-21
Run contract tests as part of CI

dtebbs:
This is a step closer to full integration tests.
- [x] launch a ganache instance and runs all the contract tests in the python client
- [x] separate contact-only tests in python client (single command to run all contract tests)
- [x] use caching to make the client and contract checks faster


client/test_contracts/__init__.py
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env python3
+
+# Copyright (c) 2015-2020 Clearmatics Technologies Ltd
```
AntoineRondelet, line 3: Could you please rebase on top of `develop` to make sure these are updated with the 2021 copyright mention and update any new header you added in this PR?

client/test_contracts/test_bls12_377_contract.py
```diff
@@ -74,53 +75,47 @@ def _b32(hex_value: str) -> bytes:
...
-if __name__ == "__main__":
-    main()
+BLS12_INSTANCE: Any = None
+
+
+class TestBLS12_377Contract(TestCase):
+
+    @staticmethod
+    def setUpClass() -> None:
+        print("Dploying BLS12_377_test.sol")
```
AntoineRondelet, line 85: Deploying

.github/workflows/onpush-checks.yml
```diff
@@ -24,7 +40,12 @@ jobs:
     - uses: actions/checkout@v1
       with:
         submodules: recursive
-    - name: Execute
+    - name: Cache pip
+      uses: actions/cache@v2
```
AntoineRondelet, line 44: Is this PR supposed to be on top of #326 ? (I can see that you are using the `cache` action here.)

dtebbs: No, this is independent of that PR.

client/test_contracts/test_bls12_377_contract.py
dtebbs: fixed typo.

client/test_contracts/__init__.py
dtebbs: should be rebased now.

AntoineRondelet: Thanks

AntoineRondelet (approved):
LGTM
Good to have more things delegated to the CI, thanks @dtebbs

PR #346, opened 2021-01-18
Fix r1cs serialization

AntoineRondelet:
Added missing scalar field characteristic to the json representation of the R1CS.

### DIFF
--- a/libzeth/serialization/r1cs_serialization.tcc
+++ b/libzeth/serialization/r1cs_serialization.tcc
@@ -153,8 +153,9 @@ std::ostream &r1cs_write_json(
 
     out_s << "{\n";
     out_s << "\"scalar_field_characteristic\":"
-          << "\"Not yet supported. Should be bigint in hexadecimal\""
-          << ",\n";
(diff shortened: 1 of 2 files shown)


libzeth/serialization/r1cs_serialization.tcc
```diff
@@ -153,8 +153,9 @@ std::ostream &r1cs_write_json(
 
     out_s << "{\n";
     out_s << "\"scalar_field_characteristic\":"
-          << "\"Not yet supported. Should be bigint in hexadecimal\""
-          << ",\n";
+          << "\"0x" +
```
dtebbs, line 156: Could pass `true` to `bitint_to_hex` instead of adding the prefix manually.

dtebbs (approved):
Another trivial thing.  LGTM, though.

libzeth/serialization/r1cs_serialization.tcc
AntoineRondelet: Thanks @dtebbs - I switched to using this flag

PR #349, opened 2021-02-04
Update zecale dispatch function signature

dtebbs:
Changes for dispatch call signature change in zecale: clearmatics/zecale#68


AntoineRondelet (approved):
LGTM

PR #361, opened 2021-03-31
Contracts cleanup

AntoineRondelet:
Refactor the solidity code to comply with the styling guide of the language and added more restrictions on function visibility and state mutation (which also allows to save gas by reading straight in the calldata instead of having the args copied).
Also removed `MerkleTreeSha256.sol` which wasn't used anymore and wasn't compiling.

Ran `slither` (https://github.com/crytic/slither) on the contracts. We're not fully compliant (still some contracts names that aren't fully in Caps, some variables that are unused in inherited contracts, uninitialized variables and ignored returns to only name a few) but the set of flags raised aren't very worrying when we dig into the code (the ignored `return` values are always `true` and come from the ERC standard etc etc). We can surely make the whole set of contracts tick all boxes but that will take a bit more time and will affect the perfs if we want to move away from low level calls and assembly - which isn't desired for now.

[EDIT] Slither doesn't seem to fully support the latest versions of Solidity (i.e. not above 0.7), so using the tool as part of the CI with the current code base (in Solidity `0.8`) isn't possible for now. Running the tool manually can be done however, with a few tricks to remove syntax not yet supported (e.g. `vk.slot`) by modifying the code (e.g. `mstore(pad, vk.slot)` -> `mstore(pad, 0x20)` to satisfy the tool and get reports for all other "Detectors")

[EDIT 2]: The full 0.7 syntax (including `.slot` and `.offset`) will soon be supported (see: https://github.com/crytic/slither/pull/831 for more details)


dtebbs:
@AntoineRondelet once we'v confirmed that #362 builds correctly, rebasing on to that should fix the CI error here.

zeth_contracts/contracts/Groth16AltBN128MixerBase_test.sol
```diff
@@ -32,40 +31,40 @@ contract Groth16AltBN128MixerBase_test is Groth16AltBN128Mixer
         public
         returns (uint256)
     {
-        return hash_public_proof_data(public_data);
+        return hashPublicProofData(public_data);
```
dtebbs, line 34: Should the method names in this file (e.g. `hash_public_proof_data_test`) also be updated?

dtebbs (approved):

zeth_contracts/contracts/Groth16AltBN128MixerBase_test.sol
AntoineRondelet: Oh yes, definitely forgot these ones. Thanks @dtebbs

AntoineRondelet:
See also: https://github.com/clearmatics/zeth/issues/335

AntoineRondelet:
This branch contains all the solidity changes. Not sure why **some** of the "file renamings" do not appear as such and are rather treated a deleted files + new files... this makes this PR extremely annoying to review. Probably did something fishy here, or maybe the set of changes on the files has reached the "diff threshold" above which git doesn't detect "renames" anymore (se e.g. https://chelseatroy.com/2020/05/09/question-how-does-git-detect-renames/), not sure.
Anyway.., @dtebbs please have a look this one when you have a moment (this PR contains #366) :)

dtebbs (approved):
LGTM

PR #365, opened 2021-04-23
Fix some generated test data

dtebbs:
Some dummy data, namely groth16 proofs (generated with known values to test serialization), has incompatible sized elements which can cause errors in some cases.

### DIFF
--- a/libzeth/tests/snarks/groth16/groth16_snark_test.cpp
+++ b/libzeth/tests/snarks/groth16/groth16_snark_test.cpp
@@ -87,20 +87,32 @@ template<typename ppT> void generate_test_data()
         -(G1::one() + G1::one()), // beta_g1 = [-2]_1
         -(G2::one() + G2::one()), // beta_g2 = [-2]_2
         libff::G1_vector<ppT>{{
-            // A_query = { [7]_1, [-3]_1, [8]_1 }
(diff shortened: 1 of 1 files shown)


AntoineRondelet:
> Some dummy data, namely groth16 proofs (generated with known values to test serialization), has incompatible sized elements which can cause errors in some cases.

I don't follow. I this PR you "extend" the `A_query` and `B_query`, what incompatibility does it fix? Likewise, what do you mean by `which can cause errors in some cases`? (why wasn't this raised on the CI by failing tests?)

dtebbs:
> > Some dummy data, namely groth16 proofs (generated with known values to test serialization), has incompatible sized elements which can cause errors in some cases.
> 
> Thanks @dtebbs - I don't quite follow though. In this PR you "extend" the `A_query` and `B_query`, what incompatibility does it fix? Likewise, what do you mean by `which can cause errors in some cases`? (why wasn't this raised on the CI by failing tests?)

In general, serialization functions don't rely on the array lengths being consistent with each other, and so it doesn't really matter how long they are.  When loading the RDD version of the proof object, indices and offsets are computed at deserialization time and stored alongside some of the elements, which requires these array lengths to be compatible with some specific primary and auxiliary input sizes.

AntoineRondelet:
Thanks @dtebbs - I see.
LGTM, merging now.

AntoineRondelet (approved):

Issue #357, opened 2021-02-25
Add a switch to enable/disable instance hashing


AntoineRondelet:
Recently, Zeth has been updated to support hashing the instance - which allows to break the linear complexity of the Snark verifier, which is ultimately required to improve the "generic" aspect of Zecale. However, the associated extra operations (on the C++ side and on the Solidity side) are not necessarily desired if Zeth is used in a standalone way. Hence, it would be nice to decide which variant to use at compile time in order to produce the appropriate binary (for use via Zecale, or in a standalone way).

AntoineRondelet:
@dtebbs let's talk about this ticket. That'd be nice to tackle it and find a reasonable approach to keep support for "standalone" Zeth while supporting aggregation via Zecale (in the current state of the code base, we don't support "standalone" Zeth anymore as we baked the Zecale integration to the code without "switches")

So here's an initial list of what's needed - but let's discuss further:
- Change the signature of the smart contract functions to make sure that an array of primary inputs can be passed (and not only 1 uint256 corresponding to the hashed instance). While we can generalize existing functions to support arrays as input, we can also refine the contract's structure and simply add new functions for the zecale integration (with their own signatures) if this is more efficient (i.e. incurs less overhead)
- Add some flags to the python and c++ code to activate/de-activate instance hashing etc
- Maybe add some compilation configuration flags to compile the project in "standalone" or "aggregated" (i.e. zecale) mode - or maybe just flags to the "prover" executable?

Note: if we want two entry points to the same Zeth contract on-chain for the "standalone" and the "aggregated" mode (which is desirable as it keeps 1 unique anonymity set), we also need to be able to verify proofs with 2 vks since the zeth statement is modified to integrate with Zecale. Hence we also need to modify the storage variables of the Mixer (to have a `_vk_standalone` and `_vk_aggregated`) etc. and call `verifyZkp` with the appropriate vk depending on the smart contract entry point (i.e. `dispatch` vs (vanilla) `mix`)

dtebbs:
If I've uderstood the issue here, the switch would only control whether to use the GGPR trick so that verification uses a single scalar multiplication, or requires a multiplication per public element.  The "standalone" terminology (which just implies operation outside of zecale) vs "aggregated" (implying operating via some batching mechanism) may be confusing (for example, we currently support "standalone" mode, we just use GGPR to make verifcation more efficient).

[...]

PR #366, opened 2021-04-23
Refined coding standards and cascaded changes to the code base

AntoineRondelet:
@dtebbs - when you have a moment, please have a look at this PR.
As the name indicates, the PR fixes some inconsistencies in our contracts code, and introduces new conventions that extend the "standard set of solidity conventions" + enforces them.

The rational is: Let's stick to solidity conventions where possible, however, small amendments are made as a way to improve smart contract dev and convey more information in the object names by following conventions exposed in the `CODING_STANDARDS` in this PR.

The TL;DR:
- Our test contracts were named as `MyContract_test.sol` -> I changed to `TestMyContract.sol` to stick to PascalCase and remain compliant with solidity standards.
- Our test functions were named inconsistently. Some were of the form `testFunction`, some of the form `functionTest`. I switched everything to the form `functionTest` (arbitrary choice, could have gone the other way around - don't have strong opinion there, so happy to switch)
- Add prefix `_` to all `private`/`internal` members. This was inconsistent throughout the code base
- Used camelCase for functions and their parameters for consistent interface using solidity standard
- Renamed our smart contracts using prefixes that convey the contract type (interface, library, "base contract", "test contract" etc).

Overall observation (for the sake of logging it):
- Our inheritances do not look very tight. There are a lot of inherited members that aren't used in child contracts. Not a big deal, but that may be something that could be improved in the future (surely not now) - though keeping tradeoffs between nice inheritance/abstraction layers and performances is necessary here.

Anyway, happy to discuss about the proposed conventions and keen to hear your opinion here.


zeth_contracts/contracts/TestBaseMixerAltBN128.sol
```diff
@@ -4,72 +4,73 @@
...
     {
-        return assembleNullifier(index, primaryInputs);
+        return _assembleNullifier(index, primaryInputs);
     }
 
     // Dummy implementation of abstract function
-    function verifyZkProof(
+    function _verifyZkProof(
         uint256[] memory /* proof */,
-        uint256[NUM_INPUTS] memory /* inputs */
```
AntoineRondelet, line 69: By refining the inheritance here it seemed that overloading `verifyZkProof` was useless in the past (because it was already defined in the former parent which was already concrete). Moreover, the signature of the function wasn't matching the signature of the abstract parent class (which tool a `uint256` for the input instead of an array of uints..)

CODING_STANDARDS.md
```diff
@@ -15,7 +15,17 @@ We adhere to the `pep8` standard, using `flake8`, `pylint` and `mypy` to catch f
...
+- Test functions must end with the suffix `Test` (e.g. `reverseBytesTest`)
+- Private/internal state variables and functions must have an underscore prefix (e.g. `_myInternalFunction`)
+- Function parameters must not be prefixed with an underscore
+- Interface names must have a capital I prefix (e.g. `IERC20`)
+- Library names must have a capital L prefix (e.g. `LPairing`)
+- Test contract names must have a `Test` prefix (e.g. `TestMyContract`)
+- Event names must be prefixed by `Log` (e.g. `LogDeposit`)
+- Contract names may not be PascalCase if using PascalCase is introducing confusions in the name (e.g. `BLS12377.sol` vs `BLS12_377.sol`). PascalCase should be used whenever possible
+
+**Note:** Some of the files of the current code base may not fully comply with the coding standards above. Old code will progressively be updated to converge towards the recommended style.
```
AntoineRondelet, line 28: More precisely, the `snake_case` convention remains broadly used in function bodies (we still have mixed styles there). The function signatures and contract's state variables/constants however should be all fine now

client/test_contracts/test_groth16_altbn128_mixer_base.py
```diff
@@ -174,17 +174,17 @@ class TestGroth16AltBN128MixerBaseContract(TestCase):
```
AntoineRondelet, line 174: TODO: Rename this python function for consistency

AntoineRondelet: Also check the python code to make sure all python functions/classes' names align with the renamed solidity contracts

zeth_contracts/contracts/BaseMixer.sol
```diff
@@ -512,8 +462,62 @@ abstract contract MixerBase is BaseMerkleTree, ERC223ReceivingContract
...
+    //
+    // where each entry entry after public output and input holds the
+    // (curve-specific) number residual bits for the corresponding 256 bit
+    // value.
+    // ======================================================================
+
+    /// Utility function to extract a full uint256 from a field element and the
+    /// n-th set of residual bits from `residual`. This function is
+    /// curve-dependent.
+    function _extractBytes32(
```
AntoineRondelet, line 504: TODO: Add to the code conventions that virtual functions should be moved at the end of the file (it's more readable and easier to see which function must be implemented to have "concrete" contracts)

zeth_contracts/contracts/BaseMixer.sol
```diff
@@ -512,8 +462,62 @@ abstract contract MixerBase is BaseMerkleTree, ERC223ReceivingContract
...
+        internal
+        pure
+        virtual
+        returns(bytes32);
+
+    /// Implementations must implement the verification algorithm of the
+    /// selected SNARK.
+    function _verifyZkProof(
+        uint256[] memory proof,
+        uint256 publicInputsHash
```
AntoineRondelet, line 518: Note to self: For https://github.com/clearmatics/zeth/issues/357 and to keep Zeth in standalone mode vs batch mode (with Zecale) -> we may want to go back to using an array here... Let's see when we tackle https://github.com/clearmatics/zeth/issues/357

CODING_STANDARDS.md
```diff
@@ -15,7 +15,18 @@ We adhere to the `pep8` standard, using `flake8`, `pylint` and `mypy` to catch f
 
 ## Solidity
 
-As of now, the solidity code does not follow consistent style conventions. Nevertheless, new solidity code must adhere to the [solidity coding standards](https://docs.soliditylang.org/en/develop/style-guide.html). Old code will progressively be updated to converge towards the recommended style.
+Solidity code must adhere to the [solidity coding standards](https://docs.soliditylang.org/en/develop/style-guide.html), with the following amendments:
+- Test functions must end with the suffix `Test` (e.g. `reverseBytesTest`)
+- Private/internal state variables and functions must have an underscore prefix (e.g. `_myInternalFunction`)
+- Virtual functions in abstract contracts must be written last and located at the end of the contract source
+- Function parameters must not be prefixed with an underscore
+- Interface names must have a capital I prefix (e.g. `IERC20`)
```
AntoineRondelet, line 23: @dtebbs - by any chance, do you remember why `IERC20` is declared as an abstract contract rather than an interface? 
See: https://github.com/clearmatics/zeth/blob/adjust-solidity-standards/zeth_contracts/contracts/Tokens.sol#L11

dtebbs: > do you remember why IERC20 is declared as an abstract contract rather than an interface?

TBH I haven't considered this, but IIRC `interface` was not available in old versions of Solidity.  My guess is that that is part of the reason.

dtebbs:
Real minor details regarding the naming, so feel free to ignore.  Probably not worth the effort of renaming, but just to give my thoughts:

> Our test contracts were named as MyContract_test.sol -> I changed to TestMyContract.sol to stick to PascalCase and remain compliant with solidity standards.
> Our test functions were named inconsistently. Some were of the form testFunction, some of the form functionTest. I switched everything to the form functionTest (arbitrary choice, could have gone the other way around - don't have strong opinion there, so happy to switch)

I'd put test at the beginning of the method names.  It sounds like an action (i.e. `testABC` does in fact test ABC), and I've see this convention used in testing frameworks to identify test methods, so it may be more familiar to people.

Conversely, (and again a real minor detail) having test class names *end* with `Test` may be more natural (instinctively a class is more like a *thing* and a method could be considered like an *action*).  Finally, having `Test` at the end is consistent with `Base` (`SomethingBase` for abstract classes sounds a bit more like "the base for building something on" whereas `BaseSomething` - which sounds a concrete "something" belonging to some "base").

> - Library names must have a capital L prefix (e.g. `LPairing`)

Although the `I` prefix is very familiar, I find `L` to be rather unintuitive.  I would probably favour either not having special names for libraries, or using `SomethingLib` (so class names can have a suffix `Base`, `Lib` or `Test`.


... or alternatively, we could consider `AbstractSomething` to match the keyword, and also `Lib` (and maybe `Test`) as suffices.  Feels a bit unnatural for `Test`, but at least everything is consistent.

Again, feel free to skip this.

dtebbs:
> Our inheritances do not look very tight. There are a lot of inherited members that aren't used in child contracts. Not a big deal, but that may be something that could be improved in the future (surely not now) - though keeping tradeoffs between nice inheritance/abstraction layers and performances is necessary here.

I'm not sure I fully understand `inherited members that aren't used in child contracts`.  Are these members that should be made private?  Or members not referenced by the base class either, in which case they can be removed?

AntoineRondelet:
Thanks @dtebbs 

> Real minor details regarding the naming, so feel free to ignore. Probably not worth the effort of renaming, but just to give my thoughts:

No problem, renaming isn't really painful once everything is consistent -> you know what patterns to search for - the real pain is when things aren't consistent ;)

> 
> > Our test contracts were named as MyContract_test.sol -> I changed to TestMyContract.sol to stick to PascalCase and remain compliant with solidity standards.
> > Our test functions were named inconsistently. Some were of the form testFunction, some of the form functionTest. I switched everything to the form functionTest (arbitrary choice, could have gone the other way around - don't have strong opinion there, so happy to switch)
> 
> I'd put test at the beginning of the method names. It sounds like an action (i.e. `testABC` does in fact test ABC), and I've see this convention used in testing frameworks to identify test methods, so it may be more familiar to people.

Good point, let's switch to that.

> Conversely, (and again a real minor detail) having test class names _end_ with `Test` may be more natural (instinctively a class is more like a _thing_ and a method could be considered like an _action_). Finally, having `Test` at the end is consistent with `Base` (`SomethingBase` for abstract classes sounds a bit more like "the base for building something on" whereas `BaseSomething` - which sounds a concrete "something" belonging to some "base").

FYI, the code base wasn't consistent w.r.t to the location of the "Base" in the name. In fact, if you check `develop` we have contracts like `<name>Base` and `BaseMerkleTree.sol`. Most of the time I find myself talking about "the base mixer" etc -> hence the name "BaseMixer" seems to better align with it (and all "Base" contracts are grouped together alphabetically which provides a better file organization IMHO).
W.r.t to the `Test<name>` vs `<name>Test` -> I'm happy to switch to `<name>Test`. Note that there seem to be no convention/consensus around this. If we check several projects like:
- Uniswap v3: https://github.com/Uniswap/uniswap-v3-core/tree/main/contracts/test
- Gnosis safe: https://github.com/gnosis/safe-contracts/tree/development/contracts/test
- Aragon: https://github.com/aragon/aragonOS/tree/next/contracts/test/tests

We see that we have a little bit of everything... The rationale behind using `Test<name>` is that (as above), the test contracts are grouped by name because they have the same prefix, and since "test" is both a name and a verb there's a bit of ambiguity here because "TestFoo" can read as "the _test contract_ for Foo" or read as "to test sthg" (i.e. "to test Foo"). Of course, a better way to group contracts would be to introduce more structure in the `contracts` folder to have `contracts/tests`, `contracts/interfaces`, `contacts/libraries` and move away from some of the prefixes, although, prefixes convey more info in the name which comes at handy in some contexts (though we can still `import as` if necessary).

> > * Library names must have a capital L prefix (e.g. `LPairing`)
> 
> Although the `I` prefix is very familiar, I find `L` to be rather unintuitive. I would probably favour either not having special names for libraries, or using `SomethingLib` (so class names can have a suffix `Base`, `Lib` or `Test`.

Many libraries are named as "lib<name>" (e.g. libfuzzer, libff, etc.) -> it seems natural to follow this prefix convention for the library names in solidity - instead of `<name>Lib.sol` - (definitely happy to switch the `L` prefix into `Lib`, that's a good idea!).

So, here's what we can do:
1. use `Lib` prefix for library files (e.g. `LibPairing.sol`)
2. use `Base` prefix for the "base" abstract classes (e.g. `BaseMixer.sol`) -> do you find "Abstract" to be more appropriate keyword here?
3. use the `testFunction` pattern for test functions, and use `ContractTest` for the "test" contracts -> Note that we do not have "1 test contract per contract".
4. I'll see if adding more levels to the `contracts` folder triggers a hell. If not, we can add subfolders to better structure the code and avoid to keep them all in the same basket (which was fine up until now, but we are starting to have quite a few contracts now) 

How does that sound to you?

AntoineRondelet:
> > Our inheritances do not look very tight. There are a lot of inherited members that aren't used in child contracts. Not a big deal, but that may be something that could be improved in the future (surely not now) - though keeping tradeoffs between nice inheritance/abstraction layers and performances is necessary here.
> 
> I'm not sure I fully understand `inherited members that aren't used in child contracts`. Are these members that should be made private? Or members not referenced by the base class either, in which case they can be removed?

Hum.. IIRC, the tool raised flags about "unused state variables" in the PGHR mixer. Such flags weren't raised for the Groth16 mixers, which seemed to imply that the inheritance structure could be refined

CODING_STANDARDS.md
AntoineRondelet: Hum... Interfaces are supported since a pretty long time: https://github.com/ethereum/solidity/blob/develop/Changelog.md#040-2016-09-08 - not sure.. anyway, will switch to an interface.

dtebbs:
> > > Our inheritances do not look very tight. There are a lot of inherited members that aren't used in child contracts. Not a big deal, but that may be something that could be improved in the future (surely not now) - though keeping tradeoffs between nice inheritance/abstraction layers and performances is necessary here.
> > 
> > 
> > I'm not sure I fully understand `inherited members that aren't used in child contracts`. Are these members that should be made private? Or members not referenced by the base class either, in which case they can be removed?
> 
> Hum.. IIRC, the tool raised flags about "unused state variables" in the PGHR mixer. Such flags weren't raised for the Groth16 mixers, which seemed to imply that the inheritance structure could be refined

Ah ... I see the point now.  Thanks.
I guess there could be a few reasons for that.  As you say, let's check the details in a future iteration and see if there are things that should be move.

zeth_contracts/contracts/BaseMixerAltBN128.sol
```diff
@@ -4,54 +4,54 @@
...
+        uint256[2] memory vkHash
     )
-        MixerBase(mk_depth, token, vk, permitted_dispatcher, vk_hash)
+        BaseMixer(mkDepth, token, vk, permittedDispatcher, vkHash)
     {
     }
 
-    /// Use MiMC7 as the Merkle tree hash function.
-    function hash(bytes32 left, bytes32 right)
+    /// Use LMiMC7 as the Merkle tree _hash function.
```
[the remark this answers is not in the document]
AntoineRondelet: Yes, thanks!

dtebbs:
> So, here's what we can do:
> 
>     1. use `Lib` prefix for library files (e.g. `LibPairing.sol`)
> 
>     2. use `Base` prefix for the "base" abstract classes (e.g. `BaseMixer.sol`) -> do you find "Abstract" to be more appropriate keyword here?
>
>     3. use the `testFunction` pattern for test functions, and use `ContractTest` for the "test" contracts -> Note that we do not have "1 test contract per contract".
> 
>     4. I'll see if adding more levels to the `contracts` folder triggers a hell. If not, we can add subfolders to better structure the code and avoid to keep them all in the same basket (which was fine up until now, but we are starting to have quite a few contracts now)
> 
> 
> How does that sound to you?

Overall that sounds good.

For (2), yes, I'd be in favour of `Abstract` as a prefix here.  I still find `BaseMixer` a bit unnatural and somewhat ambiguous (sounds like it could be a "class that mixes bases"), but if we switch to `Abstract` and `Lib` then to me it's al clear and consistent.

Re (3), tbh I could see an argument for either way.  Given that we use prefixes elsewhere I'm happy to keep the `Test...` name and align with `Lib`, `Abstract`.  Whichever you prefer.

For (4), moving tests to another direectory, like in the other components, makes a lot of sense to me if that's not a huge problem.  For other things, I'd be tempted to keep  it simple for now - dividing up by type, rather than by functionality, seems like a slightly different approach than we use elsewhere.  From that point of view it might be more natural to divide into `mimc`, `merkletree`, `alt_bn128`, `groth16` etc. but that may be overkill at the moment.

AntoineRondelet:
> Overall that sounds good.
> 
> For (2), yes, I'd be in favour of `Abstract` as a prefix here. I still find `BaseMixer` a bit unnatural and somewhat ambiguous (sounds like it could be a "class that mixes bases"), but if we switch to `Abstract` and `Lib` then to me it's al clear and consistent.

Ok sounds good let's use `Abstract` here

> Re (3), tbh I could see an argument for either way. Given that we use prefixes elsewhere I'm happy to keep the `Test...` name and align with `Lib`, `Abstract`. Whichever you prefer.

I have no strong preferences as per my comments above. In the absence of justification to change, let's stick to what we have now and use `Test` as a prefix. We can change later if this turns out to be problematic (though I doubt it)

> From that point of view it might be more natural to divide into `mimc`, `merkletree`, `alt_bn128`, `groth16` etc. but that may be overkill at the moment.

Hum... such a divide looks pretty speculative as it seems to make some assumption about the future state of the code base. The current st of contract wouldn't benefit from such a granular divide. Of course, as the code base evolves, new conventions become more appropriate and new structure emerge. So, to keep things focused now, let's postpone this folder divide for later. I'll move the tests in their folder, and we'll have a handful of contracts in the same basket - which is perfectly tractable for now - especially with the newly adopted conventions on the names

AntoineRondelet:
> I'll move the tests in their folder, and we'll have a handful of contracts in the same basket - which is perfectly tractable for now - especially with the newly adopted conventions on the names

This will be one for another PR because this requires to touch the python code (beyond renaming) in order to modify the path to the contracts to deploy in `mock.deploy_contract` (especially in `zeth.utils.get_contracts_dir()`)

- https://github.com/clearmatics/zeth/blob/develop/client/test_commands/mock.py#L98-L99
- https://github.com/clearmatics/zeth/blob/28d901e9d1f909ace4385a79af32df50f5355c65/client/zeth/core/utils.py#L255-L258

CODING_STANDARDS.md
```diff
@@ -15,7 +15,19 @@ We adhere to the `pep8` standard, using `flake8`, `pylint` and `mypy` to catch f
...
 ## Solidity
 
-As of now, the solidity code does not follow consistent style conventions. Nevertheless, new solidity code must adhere to the [solidity coding standards](https://docs.soliditylang.org/en/develop/style-guide.html). Old code will progressively be updated to converge towards the recommended style.
+Solidity code must adhere to the [solidity coding standards](https://docs.soliditylang.org/en/develop/style-guide.html), with the following amendments:
+- Test functions must have a `test` prefix (e.g. `testReverseBytes`)
+- Private/internal state variables and functions must have an underscore prefix (e.g. `_myInternalFunction`)
+- The order in which contract members are written must driven by their scope (i.e. `public/external` functions must appear first in the contract code, `internal/private` functions must appear last). Additionally, `virtual` functions in abstract contracts must be written last in their visibility group.
+- Function parameters must not be prefixed with an underscore
+- Interface names must have a capital I prefix (e.g. `IERC20`)
+- Library names must have a capital L prefix (e.g. `LibPairing`)
```
dtebbs, line 24:
```suggestion
- Library names must have a `Lib` prefix (e.g. `LibPairing`)
```

dtebbs (approved):
LGTM with one comment.

CODING_STANDARDS.md
AntoineRondelet: Great catch, thanks @dtebbs

PR #377, opened 2021-05-12
Zeth tool (depends on #376)

dtebbs:
Collect utils for checking and manipulating proofs (useful for debugging) into a single `zeth-tool`, sharing infrastructure with mpc tools.


CMakeLists.txt
```diff
@@ -311,11 +311,12 @@ endif()
 
 # Add all local subdirecetories
 add_subdirectory(libzeth)
+add_subdirectory(libtool)
 
 # If zeth is being used as a dependency, skip the tools build
 if ("${IS_ZETH_PARENT}")
   add_subdirectory(prover_server)
-  add_subdirectory(verifier)
+  add_subdirectory(zeth_tool)
```
AntoineRondelet, line 319: What about naming it `zethctl`?

libtool/run.tcc
```diff
@@ -0,0 +1,149 @@
...
+
+template<typename GlobalOptionsT>
+void print_usage(
+    const char *const *argv,
+    const po::options_description &options,
+    const std::map<std::string, subcommand<GlobalOptionsT> *> &subcommands)
+{
+    std::cout << "Usage:\n"
+              << "  " << argv[0]
+              << " [<options>] <command> <command-arguments> ...\n\n"
```
AntoineRondelet, line 24: Why aren't the options after the command: `zeth_tool verify <my_options>` (`<my options>` being options of `zeth_tool` or options of `verify` (i.e the command))?

libtool/run.tcc
```diff
@@ -0,0 +1,149 @@
...
+              << options;
+
+    std::cout << "\nCommands:\n";
+
+    // list_commands(commands);
+    using entry_t = std::pair<std::string, subcommand<GlobalOptionsT> *>;
+    const size_t cmd_name_padded =
+        4 + std::max_element(
```
AntoineRondelet, line 32: Where is 4 coming from?

mpc_tools/mpc_phase2/mpc_client.cpp
```diff
@@ -28,10 +28,10 @@ void zeth_protoboard(libsnark::protoboard<Field> &pb)
...
 int main(int argc, char **argv)
 {
-    const std::map<std::string, subcommand *> commands{
+    const std::map<std::string, mpc_subcommand *> commands{
         {"phase2-contribute", mpc_phase2_contribute_cmd},
         {"phase2-verify-transcript", mpc_phase2_verify_transcript_cmd},
         {"create-keypair", mpc_create_keypair_cmd},
     };
-    return mpc_main(argc, argv, commands, zeth_protoboard);
+    return mpc_main(commands, zeth_protoboard, argc, argv);
```
AntoineRondelet, line 36: What's the rationale in the argument order switch here?

mpc_tools/mpc_phase2/cli/mpc_subcommand.hpp
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2015-2021 Clearmatics Technologies Ltd
+//
+// SPDX-License-Identifier: LGPL-3.0+
+
+#ifndef __ZETH_MPC_CLI_MPC_SUBCOMMAND_HPP__
```
AntoineRondelet, line 5: `__ZETH_MPC_CLI_MPC_SUBCOMMAND_HPP__` -> `__MPC_TOOLS_MPC_PHASE2_CLI_MPC_SUBCOMMAND_HPP__` or `__MPC_TOOLS_PHASE2_CLI_MPC_SUBCOMMAND_HPP__` ?

CMakeLists.txt
dtebbs: I'm not sure what the `ctl` part means.  I think it should be something pretty intuitive.

libtool/run.tcc
dtebbs: This part refers to the global options, which appear before the command name.  The arguments for specific commands (`<command-arguments>`) are described in the usage for that command.

mpc_tools/mpc_phase2/mpc_client.cpp
dtebbs: This matches the order of the `libtool::run` call.  The first few parameters specialise the command, and the later args are per-invocation.  (You might think of the mpc_client `main` function as being the partial application of those first few arguments to `mpc_main`).

mpc_tools/mpc_phase2/cli/mpc_subcommand.hpp
dtebbs: Updated.  Thanks.

libtool/run.tcc
dtebbs: It's in order to ensure padding between the command name and the description.  I've added a comment.

CMakeLists.txt
AntoineRondelet: > I'm not sure what the ctl part means

Ah yes, ctl = control
(else `zethcli` or `zeth-cli`?)

`zeth-tool` is fine, simply not a huge fan of the "tool" suffix here - i.e. most of the time we talk about a command line interface, not a "tool" which sounds rather very generic.

AntoineRondelet: Ah I just saw your comment after pushing mine above (apparently needed to update the UI), yeah, that's what I thought (the collision with the client's cli)

dtebbs: I had another look and `zeth_helper` should now be `zeth-helper` in the latest commit.

AntoineRondelet: > On balance, I'd probably prefer to avoid the cli in the name. It always seems superfluous (you know it's the cli if you are invoking it from the shell) 

Yes that's a good point.

> and even more generic than tool or util.

I think this is arguable - to me `cli` and `tool` are as generic tbh, but yes that's right `cli` isn't really a good prefix too I agree.

> The logic with adding a suffix here (instead of in the python client) was this the python client zeth is the main entry point for most administrator and client wallet operations, whereas this new tool is for more detailed operations.

Yes, I was surely not suggesting to add `tool` as prefix to the zeth wallet cli, I perfectly see why you would add it here.
The general point is that we start to have quite a few CLIs: the cpp one (the "tool"), the wallet one ("zeth") which in fact has more functionalities than just the "wallet", the helper one etc. I think a little bit of renaming wouldn't hurt here.
From what I can see, the "tool" (i.e. `zeth-tool`) is used to manipulate the proof objects with the subcommands:
```
        {"verify", verify_cmd},
        {"prove", prove_cmd},
        {"dump-proof", dump_proof_cmd},
```
So maybe we can rename the executable `zeth-proof` and the commands as `zeth-proof generate`, `zeth-proof verify`, `zeth-proof dump`. Of course, this won't work anymore if the tool is extended to carry out non-proof related work.

If we are short of names for now we can stick to "zeth-tool" no big deal at this stage really, but I'd suggest we add a README in https://github.com/clearmatics/zeth/tree/zeth-tool/zeth_tool to better explain the role of the executable.

> Regarding the zeth_helper, I've tried to use - consistently for command names, but IIRC python would not allow - for some reason. Although there may be a different way to add that command - I'll look into it.
>> I had another look and zeth_helper should now be zeth-helper in the latest commit.

Thanks. Here again no huge deal, but in general I prefer to avoid separators `_` and `-` in the names of the commands. Some commands use `_` some use `-` and I find it confusing because there aren't conventions about that (so you end up trying with an `-` and then figure out that it should be `_` after getting an error the first time. Clearer to avoid it all together when the name isn't too ugly)

dtebbs: I agree `tool` is somewhat generic.  However, it does at least imply something a bit beyond the main functionality.  Not necessariily central to the primary operation, but for advanced or more special-case operations.

While the current functionality is mainly around proofs, etc, but including `proof` in th name may be a bit too specific.  The intention is that any other operations could be added here.  For example (as discussed in the past) it could make sense to have this tool perform the one-shot dummy keypair generation, rather than relying on the server doing it automatically if an existing keypair file does not exist.  As another example, if this tool had existed earlier, I would have added functionality to extract the PK and VK from a keypair file, to generate test data.  I'm expecting it to build up this kind of functionality, and possibly eventually be invoked by the main client tools where that makes sense.

I've added a brief readme to try and capture the intent.  However, I still can't think of anything better than `zeth-tool` for the name.

zeth_tool/dump_proof_cmd.cpp
```diff
@@ -0,0 +1,83 @@
...
+        std::cout << "Usage:\n  " << argv0 << " dump-proof [proof_file]\n";
+    }
+
+    std::string proof_file;
+};
+
+} // namespace commands
+
+zeth_subcommand *dump_proof_cmd =
+    new commands::dump_proof_cmd("dump-proof", "Print an existinng proof");
```
AntoineRondelet, line 81: `existinng` -> `existing`

zeth_tool/main.cpp
```diff
@@ -0,0 +1,47 @@
...
+};
+
+int main(int argc, char **argv)
+{
+    // Create command structures
+    std::map<std::string, zeth_subcommand *> commands{
+        {"verify", verify_cmd},
+        {"prove", prove_cmd},
+        {"dump-proof", dump_proof_cmd},
+    };
```
AntoineRondelet, line 40: If I try invalid command names, I get a segmentation faults e.g.
```bash
17:21:56 ~/D/t/z/build ❯❯❯ ./zeth_tool/zeth-tool verif                            ✘ 1
[1]    25886 segmentation fault (core dumped)  ./zeth_tool/zeth-tool verif
17:22:11 ~/D/t/z/build ❯❯❯ ./zeth_tool/zeth-tool generate                       ✘ 139
[1]    25935 segmentation fault (core dumped)  ./zeth_tool/zeth-tool generate
```

That would be preferable to fail safe here and simply display a message `command not found` and fallback to the help menu.

libtool/run.tcc
AntoineRondelet: I see, thanks @dtebbs 

Here's a detail w.r.t the usage message. We currently have:
```
$ zeth
Usage: zeth [OPTIONS] COMMAND [ARGS]...

Options:
  --eth-network TEXT         Ethereum RPC endpoint, network or config file
                             (default: 'eth-network' if it exists, otherwise
                             'ganache')
  --prover-server TEXT       Prover server endpoint (default=localhost:50051)
  --prover-config-file TEXT  Prover config file (default=prover-config.cache)
  --instance-file TEXT       Instance file (default=zeth-instance)
  --address-file TEXT        Zeth zecret address file (default=zeth-
                             address.priv)
  --wallet-dir TEXT          Wallet directory (default=./wallet)
  --help                     Show this message and exit.

Commands:
  help*                 Print help and exit
  deploy                Deploy the zeth contracts and record the...
  gen-address           Generate a new Zeth secret key and public address
  get-verification-key  Command help text.
  ls-commits            List all commitments in the Zeth mixer contract
  ls-notes              List the set of notes owned by this wallet
  mix                   Generic mix function
  sync                  Attempt to retrieve new notes for the key in...
  wait                  Wait for a mix transaction and dump all log data.
Error: no command specified

$ zeth-tool                         ✘ 1
error: no command specified
Usage:
  ./build/zeth_tool/zeth-tool [<options>] <command> <command-arguments> ...

Global options:
  -h [ --help ]         This help
  -v [ --verbose ]      Verbose output

Commands:
  dump-proof    Print an existinng proof
  prove         Generate proof given proving key and assignment
  verify        Verify proof against verification key and primary input


$ zeth-helper
Usage: zeth-helper [OPTIONS] COMMAND [ARGS]...

Options:
  --eth-network TEXT  Ethereum RPC endpoint, network or config file (default:
                      'eth-network' if it exists, otherwise 'ganache')
  --help              Show this message and exit.

Commands:
  help*                     Print help and exit
  eth-fund                  Fund an address.
  eth-gen-address           Locally generate a new Ethereum private key and...
  eth-gen-network-config    Generate a network config file.
  eth-get-balance           Command to get the balance of specific...
  eth-get-contract-address  Extract the address from a contract instance...
  eth-send                  Send Ether from the local eth-addr to a...
  token-approve             Approve the mixer to spend some amount of...
Error: no command specified
```


As such, I'd change the "usage string" of `zeth-tool` from `zeth-tool [<options>] <command> <command-arguments> ...` to `zeth-tool [OPTIONS] COMMAND [ARGS]...` for the sake of interface consistency across our various CLIs

zeth_tool/prove_cmd.cpp
```diff
@@ -0,0 +1,130 @@
...
+        {
+            std::ifstream in_s = libtool::open_binary_input_file(pk_file);
+            snarkT::proving_key_read_bytes(proving_key, in_s);
+        }
+
+        libsnark::r1cs_primary_input<libff::Fr<ppT>> primary;
+        libsnark::r1cs_auxiliary_input<libff::Fr<ppT>> auxiliary;
+        {
+            std::ifstream in_s =
+                libtool::open_binary_input_file(assignment_file);
```
AntoineRondelet, line 45: Do I understand correctly that the only way for me to generate an assignment file for now is via https://github.com/clearmatics/zeth/blob/master/prover_server/prover_server.cpp#L116, and thus by running: `prover_server --assignment-output <filename>`?

CMakeLists.txt
AntoineRondelet: > I've added a brief readme to try and capture the intent.

Great, thanks @dtebbs 

> However, I still can't think of anything better than zeth-tool for the name.

No big deal, the README documents the CLI's purpose. We can rename the "tool" if we find a better name in the future.

libtool/run.tcc
```diff
@@ -0,0 +1,151 @@
...
+    GlobalOptionsT &options,
+    const std::map<std::string, subcommand<GlobalOptionsT> *> &subcommands,
+    int argc,
+    char **argv)
+{
+    po::options_description global("Global options");
+    po::options_description all("");
+
+    // Default --help option
+    global.add_options()("help,h", "This help");
```
AntoineRondelet, line 81: `"This help"` -> `"Show this message and exit."`

mpc_tools/mpc_phase2/mpc_client.cpp
AntoineRondelet: I see, thanks.

zeth_tool/verify_cmd.cpp
```diff
@@ -0,0 +1,119 @@
...
+        boost::program_options::options_description &options,
+        boost::program_options::options_description &all_options,
+        boost::program_options::positional_options_description &pos) override
+    {
+        base_class::initialize_suboptions(options, all_options, pos);
+
+        all_options.add(options).add_options()(
+            "vk_file", po::value<std::string>(), "Verification key file");
+        all_options.add_options()(
+            "primary_input_file", po::value<std::string>(), "Proof file");
```
AntoineRondelet, line 72: `"Proof file"` -> `"Primary input file"`

zeth_tool/dump_proof_cmd.cpp
dtebbs: fixed.  thanks.

zeth_tool/main.cpp
dtebbs: fixed.  thanks.

zeth_tool/prove_cmd.cpp
dtebbs: Changed the usage string.

(Yes - currently the only way to generate the assignment is via the prover_server.)

libtool/run.tcc
dtebbs: Change flag description text

zeth_tool/verify_cmd.cpp
dtebbs: fixed.  thanks.

zeth_tool/prove_cmd.cpp
AntoineRondelet: Got it, thanks

zeth_tool/main.cpp
AntoineRondelet: Great, thanks!

AntoineRondelet (approved):
LGTM, thanks @dtebbs

PR #378, opened 2021-05-21
Tool changes related to joinsplit inputs and outputs (depends on #379)

dtebbs:


client/zeth/core/mixer_client.py
```diff
@@ -826,21 +810,31 @@ def _trap_r_randomness() -> str:
 
 def _compute_rho_i(phi: bytes, hsig: bytes, i: int) -> bytes:
     """
-    Returns rho_i = blake2s(0 || i || 10 || [phi]_252 || hsig)
+    Returns rho_i = blake2s(0 || (i..i) || 10 || [phi_truncated] || hsig)
```
AntoineRondelet, line 813: What is `i..i`? Isn't it `i`? (i.e. if `1..5` -> 1,2,3,4,5, then `i..i` -> i). What do I miss?

dtebbs: This is supposed to represent the bits in the binary representation of i.  I've adjusted the comments, so hopefully it's clearer now.  Thanks.

AntoineRondelet: I see, thanks that's clearer now

AntoineRondelet (approved):
LGTM, thanks @dtebbs

PR #382, opened 2021-06-17
Use latest libff optimizations

dtebbs:


libzeth/tests/snarks/groth16/groth16_snark_test.cpp
```diff
@@ -81,7 +81,7 @@ template<typename ppT> void generate_test_data()
 
     // Proving key
     const typename groth16::proving_key pk(
-        G1::one(),                // alpha_g1 = [1]_1
+        G1(G1::one()),            // alpha_g1 = [1]_1
```
AntoineRondelet, line 84: Why do you need to surround `G1:one()` here?

zeth_tool/prove_cmd.cpp
```diff
@@ -71,7 +75,8 @@ class prove_cmd : public generic_subcommand<prove_cmd>
         options.add_options()(
             "primary_inputs,p",
             po::value<uint16_t>(),
-            "Number of primary inputs (default: 1)");
+            "Number of primary inputs (default: 1)")(
+            "profile,p", "Enable profiling output");
```
AntoineRondelet, line 79: Don't you have a clash in the "short flags" names here? (i.e. `p` is used for both the short flag of `profiling` and `primary inputs`)?

libzeth/tests/snarks/groth16/groth16_snark_test.cpp
dtebbs: This is required because the constructor of this type (`libsnark::r1cs_gg_ppzksnark_proving_key`) only accepts rvalue refs, but `G1::one` now returns a const ref instead of a value.

(AFAICT there is no advantage to rval refs in that constructor, but I didn't want to start chaning the constructor signatures at this point)

AntoineRondelet: Ah, I see. Thanks.

zeth_tool/prove_cmd.cpp
dtebbs: Ah yes.  Thanks.  Switched to `r` for profile.

AntoineRondelet: Thanks

AntoineRondelet (approved):
LGTM, thanks @dtebbs

Issue #314, opened 2020-11-23
Clean up `scripts` folder


AntoineRondelet:
As described in https://github.com/clearmatics/zeth/pull/309#discussion_r528665321, introducing more structure in the `scripts` folder would better organize our ever growing set of scripts.

PR #385, opened 2021-07-01
Scripts to generate large circuits, keys and assignments

dtebbs:
Scripts and tools related to generating zeth witness data for various JSIN/OUT values, 
- [x] `zeth-tool split-keypair` command
- [x] fixes and scripts for autmating witness data generation
- [x] improvements to some build scripts, support for building on ec2 instances
- [x] some verbose client output related to tx size

(Could potentially drop some of the commits directly related to data generation.  For now, leaving everything for discussion.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.15)
+cmake_minimum_required(VERSION 3.13)
 
 # Change the compiler BEFORE the first `project()` command to avoid an infinite loop.
 # See: https://public.kitware.com/pipermail/cmake/2009-November/033133.html
--- a/scripts/ci_utils.sh
+++ b/scripts/ci_utils.sh
@@ -147,6 +147,13 @@ function cpp_build_setup() {
         if (which apk) ; then
             # Packages already available in Docker build
             echo -n             # null op required for syntax
+        elif (which yum) ; then
+            sudo yum groupinstall -y "Development Tools"
+            sudo yum install -y \
+                 openssl openssl-devel \
+                 gmp-devel procps-devel cmake3 \
+                 python3 python3-devel \
+                 boost-devel
         else
             sudo apt install \
                  libboost-dev \
--- a/scripts/ci_utils.sh
+++ b/scripts/ci_utils.sh
@@ -97,13 +97,14 @@ function prover_server_is_active() {
     zeth get-verification-key
 }
 
+# 1 - prover server flags
 function prover_server_start() {
     # Requires the client env (for prover_server_is_active)
     . client/env/bin/activate
     pushd build
 
     server_start \
-        ./prover_server/prover_server \
+        "./prover_server/prover_server $1" \
         prover_server_is_active \
         prover_server.pid \
         prover_server.stdout
--- a/scripts/ci_utils.sh
+++ b/scripts/ci_utils.sh
@@ -3,6 +3,8 @@
 # All functions expect to be executed the root directory of the repository, and
 # will exit with this as the current directory.
 
+./scripts/build_utils.sh
+
 # Launch a server in the background and wait for it to be ready, recording the
 # pid in a file.
 #
@@ -48,6 +50,8 @@ function server_stop() {
 #
 
 function ganache_setup() {
+    assert_init_platform
+
     if [ "${platform}" == "Linux" ] ; then
         if (which apk) ; then
             apk add --update npm
@@ -117,53 +121,4 @@ function prover_server_stop() {
     pushd build
     server_stop prover_server prover_server.pid
     popd # build
-}
-
-#
-# DEPENDENCIES
-#
-
-function cpp_build_setup() {
-    # Extra deps for native builds
-
-    if [ "${platform}" == "Darwin" ] ; then
-        # Some of these commands can fail (if packages are already installed,
-        # etc), hence the `|| echo`.
-        brew update || echo
-        brew install \
-             gmp \
-             grpc \
-             protobuf \
-             boost \
-             openssl \
-             cmake \
-             libtool \
-             autoconf \
-             automake \
-             || echo
-    fi
-
-    if [ "${platform}" == "Linux" ] ; then
-        if (which apk) ; then
-            # Packages already available in Docker build
-            echo -n             # null op required for syntax
-        elif (which yum) ; then
-            sudo yum groupinstall -y "Development Tools"
-            sudo yum install -y \
-                 openssl openssl-devel \
-                 gmp-devel procps-devel cmake3 \
-                 python3 python3-devel \
-                 boost-devel
-        else
-            sudo apt install \
-                 libboost-dev \
-                 libboost-system-dev \
-                 libboost-filesystem-dev \
-                 libboost-program-options-dev \
-                 libgmp-dev \
-                 libprocps-dev \
-                 libxslt1-dev \
-                 pkg-config
-        fi
-    fi
 }
--- a/scripts/ec2_setup.sh
+++ b/scripts/ec2_setup.sh
@@ -1,0 +1,66 @@
+#!/usr/bin/env bash
+
+# Installs dependencies and runs a build of all components required to execute
+# the zeth pipeline. Specialized for ec2 instances using the 'yum' package
+# manager.
+
+if ! (which yum) ; then
+    echo ERROR: yum package manager not found.
+    exit 1
+fi
+
+set -e
+set -x
+
+#
+# Dependencies
+#
+
+# Packages
+sudo yum groupinstall -y "Development Tools"
+sudo yum install -y \
+     openssl openssl-devel \
+     gmp-devel procps-devel cmake3 \
+     python3 python3-devel \
+     boost-devel \
+     emacs-nox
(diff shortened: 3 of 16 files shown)


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.15)
+cmake_minimum_required(VERSION 3.13)
```
AntoineRondelet, line 1: Why decrease the version here?

scripts/ci_utils.sh
```diff
@@ -3,6 +3,22 @@
 # All functions expect to be executed the root directory of the repository, and
 # will exit with this as the current directory.
 
+
+# Init some global variables related to the platform. Some other functions may
+# expect this to be called before they invoked.
```
AntoineRondelet, line 8: they are invoked

scripts/ci_utils.sh
```diff
@@ -3,6 +3,22 @@
...
+function init_platform() {
+    platform=`uname`
+    echo platform=${platform}
+}
+
+# Assert that init_platform has been called
+function assert_init_platform() {
+    if [ "" == "${platform}" ] ; then
```
AntoineRondelet, line 16: For consistency, in such a conditional (where a variable value is compared to a constant or an hardcoded value) the variable should be on the left, the value/constant it is compared to should be on the right: `[ "${platform}" == "" ]`

scripts/ci_utils.sh
```diff
@@ -146,6 +167,13 @@ function cpp_build_setup() {
         if (which apk) ; then
             # Packages already available in Docker build
             echo -n             # null op required for syntax
+        elif (which yum) ; then
+            sudo yum groupinstall -y "Development Tools"
```
AntoineRondelet, line 171: This seems to be useful for experimenting on an Amazon Linux instance. I.e. AFAIK we do not use RHEL-type OSes on the CI. If this is indeed used outside of the scope of the CI, we probably want to re-arrange the scripts to move this function in another file than this "ci_utils" script.

scripts/ec2_setup.sh
```diff
@@ -0,0 +1,57 @@
...
+
+set -e
+set -x
+
+#
+# Dependencies
+#
+
+# Packages
+sudo yum groupinstall -y "Development Tools"
```
AntoineRondelet, line 11: This implies that the EC2 instance is Amazon Linux (or RHEL or a derivative). Running such script on a Debian/Ubuntu EC2 instance will certainly fail with a `"Command not found: yum"` type error. May be worth adding a comment in the script, e.g.
```
#
# Dependencies (assuming `yum` as a package manager)
#
```

scripts/install_grpc
```diff
@@ -3,75 +3,73 @@
 # This script is only intended to run on the CI machines. Not for local
 # development.
 
-# Expect 3 arguments, all non-empty.
-if [ "$#" -ne 3 ] || [ "" == "$1" ] || [ "" == "$2" ] || [ "" == "$3" ] ; then
+# Expect 2 arguments.
+if [ "" == "$1" ] || [ "" == "$2" ] ; then
```
AntoineRondelet, line 7: To be consistent, let's switch to:
```
if [ "$1" == "" ] || [ "$2" == "" ] ; then
```

scripts/prover_profile.sh
```diff
@@ -0,0 +1,24 @@
+
+CIRCUIT=2_2
```
AntoineRondelet, line 2: Let's document this to say that the variable CIRCUIT takes a value of the form `<JSIN>_<JSOUT>`

scripts/prover_profile.sh
```diff
@@ -0,0 +1,24 @@
...
+
+ZETH_TOOL=./zeth_tool/zeth-tool
+PK=~/dev/neodizk/test_data/zeth/zeth_${CIRCUIT}_pk.bin
+VK=~/dev/neodizk/test_data/zeth/zeth_${CIRCUIT}_vk.bin
```
AntoineRondelet, line 6: Why do we have these paths here? This implies a given structure on the user's machine. I'd rather use the values in https://github.com/clearmatics/zeth/blob/master/setup_env.sh when possible, or simply document these values. e.g.
```
# VK=~/dev/neodizk/test_data/zeth/zeth_${CIRCUIT}_vk.bin

NEODIZK=<your_path_to_neodizk> # To use this script please, close neodizk https://github.com/clearmatics/neodizk and point to its location here.
VK=$NEODIZK/test_data/zeth/zeth_${CIRCUIT}_vk.bin
```

zeth_tool/split_keypair_cmd.cpp
```diff
@@ -0,0 +1,109 @@
...
+        options.add_options()(
+            "vk-file,v",
+            po::value<std::string>(),
+            "Verification key file (optional)")(
+            "pk-file,p",
+            po::value<std::string>(),
+            "Proving key file (optional)");
+
+        all_options.add(options).add_options()(
+            "keypair_file,k", po::value<std::string>(), "Keypair file");
```
AntoineRondelet, line 69: Use of snake_case in `keypair_file` is inconsistent with the use of `-` above (i.e. `pk-file` and `vk-file`), so let's switch to `keypair-file`

AntoineRondelet (commented):
See comments in the PR.
As discussed off line, I think it's perfectly fine to add the scripts related to the experimentation with neodizk on `zeth/master`. However, I would isolate them in a separate sub-folder under the `scripts` folder (this is relevant to https://github.com/clearmatics/zeth/issues/314). As a matter of fact, moving them in a separate subfolder would allow us to add a README `scrips/neodizk/README.md` in which we can put more color about the usefulness of these scripts, i.e. explain which OS is assumed for the EC2 instance, what the user setup should look like (where the neodizk is assumed to be cloned on the machine) etc etc.

CMakeLists.txt
dtebbs: This is for compatibility with platforms which currently don't supply v 3.15.

zeth_tool/split_keypair_cmd.cpp
dtebbs: renamed the argument key

scripts/prover_profile.sh
dtebbs: Changed as part of the refactor above.  These values are now more configurable, falling back to defaults.

scripts/prover_profile.sh
dtebbs: Refactored the script and added comments to make it slightly more generic.  This has changed as part of that refactor.

scripts/install_grpc
dtebbs: Changed value order

scripts/ec2_setup.sh
dtebbs: added some comments and a check in the script

scripts/ci_utils.sh
dtebbs: moved to a general `build_utils.sh` script.

scripts/ci_utils.sh
dtebbs: switched value order

scripts/ci_utils.sh
dtebbs: fixed typo

CMakeLists.txt
AntoineRondelet: Got it

scripts/prover_profile.sh
```diff
@@ -0,0 +1,48 @@
...
+#   ${CIRCUIT}_pk.bin
+#   ${CIRCUIT}_vk.bin
+#   ${CIRCUIT}_assignment.bin
+#   ${CIRCUIT}_primary.bin
+# in the directory ${DATA_DIR} (set on command line or env).
+#
+# The zeth-tool binary is expected to be in the path, or at ${ZETH_TOOL}.
+
+# Set default DATA_DIR
+[ "${DATA_DIR}" == "" ] && DATA_DIR=~/dev/neodizk/test_data/zeth
```
AntoineRondelet, line 15: As per my previous comment https://github.com/clearmatics/zeth/pull/385#pullrequestreview-731005028, if this script is to be mostly used in the context of neodizk<>zeth (which is suggested by the default value of `DATA_DIR`), I'd add more color here on the value of DATA_DIR. This can either be done by adding a line of comment above, or by moving this script in a `scripts/neodizk` folder and providing the necessary context in a README. If we want this script to be rather generic and used in several contexts, then I'd refrain from pointing to a location under `neodizk` (which is rather confusing) and would rather use a default path that is under `zeth/` and maybe add a comment to explain why we may want to point to a `neodizk` location in some circumstances.

dtebbs: > As per my previous comment #385 (review), if this script is to be mostly used in the context of neodizk<>zeth 

This script profiles the code in this repository, so no need to reference external projects.   I've removed the default values.

AntoineRondelet: Thanks, I think this is cleaner that way, else you are left wondering why the default value points to a `neodizk` location. If we want to use a default value that's fine but it should rather point to the `testdata` folder of zeth, though that's fine like this, users will set the env var manually to fit their config.

AntoineRondelet (approved):
LGTM, thanks @dtebbs

PR #388, opened 2021-08-04
Update libsnark

dtebbs:
Pending:
- [x] clearmatics/libsnark#20

(once the above is merged, this PR can be marked as ready)

PR #389, opened 2021-08-10
Update web3 and pylint packages in client (depends on #388)

dtebbs:
web3.py update did not really eliminate as much code as I had hoped.  They provide some better interfaces for parsing log data, so we avoid importing a few undocumented functions, but unless I'm missing something it's still not easy to iterate through a list of logs, filtering them by name (without jupming through some hoops or making multiple requests to the node).

A new minor version of pylint (which got dragged in when I relaxed some version conditions as part of the web3.py udpate) seems a bit stricter about import formats.


dtebbs:
Changes are complete (removing WIP and requesting review) , but keeping as draft until other PRs are merged.

AntoineRondelet (approved):
LGTM, thanks @dtebbs

Issue #295, opened 2020-10-08
Rely on `estimateGas` for all contract calls


AntoineRondelet:
For now, a few commands of the `zeth` CLI still use hardcoded gas value (see: `zeth mix` -> https://github.com/clearmatics/zeth/blob/develop/client/zeth/core/mixer_client.py#L420 and https://github.com/clearmatics/zeth/blob/develop/client/zeth/core/contracts.py#L281 for instance), while other commands (such as `zeth_helper token-approve` -> https://github.com/clearmatics/zeth/blob/develop/client/zeth/helper/token_approve.py#L45) do not pass the optional gas value and rather rely on `estimateGas`. That would be nice to make this uniform across the various commands, and rely on estimateGas when possible.

AntoineRondelet:
This ticket has been addressed in #390. Closing as a consequence.

PR #390, opened 2021-08-11
Remove hardcoded gas (depends on #389)

dtebbs:
Related to #295 #281 clearmatics/autonity#681
Checked against autonity using network config:
```
{"name": "bakerloo", "endpoint": "https://rpc3.bakerloo.autonity.network:8545"}
```


dtebbs:
NOTE: May need to check that this doesn't break anything w.r.t. the `estimateGas` API call (changes to this call, including its signature).

AntoineRondelet (approved):
LGTM, great thanks @dtebbs, did some tests on Rinkeby, all good there.

Issue #387, opened 2021-07-20
bug in sha256_ethereum.tcc | get_hash


kaxxa123:
Hello there is a little bug in this code:
https://github.com/clearmatics/zeth/blob/master/libzeth/circuits/sha256/sha256_ethereum.tcc

Function:
template<typename FieldT>
libff::bit_vector sha256_ethereum<FieldT>::get_hash(
    const libff::bit_vector &input)

This function is missing the ZERO parameter
   sha256_ethereum<FieldT> eth_hasher(
        pb, input_block, output_variable, "eth_hasher_gadget");

This causes a compile time error so its not serious.

I fixed this as follows:

template<typename FieldT>
libff::bit_vector sha256_ethereum<FieldT>::get_hash(
    const libff::bit_vector &input)
{
    libsnark::protoboard<FieldT> pb;

    libsnark::pb_variable<FieldT> ZERO;
    ZERO.allocate(pb, "ZERO");

    libsnark::block_variable<FieldT> input_block(
        pb, libsnark::SHA256_block_size, "input_block");
    libsnark::digest_variable<FieldT> output_variable(
        pb, libsnark::SHA256_digest_size, "output_variable");
    sha256_ethereum<FieldT> eth_hasher(
        pb, ZERO, input_block, output_variable, "eth_hasher_gadget");

    pb.val(ZERO) = FieldT::zero();
    input_block.generate_r1cs_witness(input);
    eth_hasher.generate_r1cs_witness();

    return output_variable.get_digest();
}

dtebbs:
Thanks for raising this @kaxxa123, and for the code snippet.  I guess that gadget has not been used for a while and needs some attention (and a test).  Feel free to submit a pull request with your fix.

AntoineRondelet:
Thanks @kaxxa123! Good catch, the function signature is not satisfied indeed justifying the compilation error.

Rather than quick fixing this and add ZERO back, we should probably bring the sha256 hasher interface in sync with the blake one (which doesn't use ZERO), see:
https://github.com/clearmatics/zeth/blob/8247fc3df9d0025000196293e121330440178ec0/libzeth/circuits/circuit_utils.tcc#L51-L64

Removing ZERO from the arguments of the sha constructor begs for a refactoring of this "legacy" function that uses ONE and ZERO. So we need to get rid of the useless directives at the top of the file and use FieldT::one(), FieldT::zero() apropriately here:
[...]

PR #391, opened 2021-08-11
Fix sha256 gadget (depends on #390)

dtebbs:
- [x] remove some uses of ZERO in bit-string variables
- [x] test for `sha256_ethereum::get_hash()`
- [x] remove ZERO in `sha256_ethereum` constructor
- [x] factor out some bit-string code (which is no longer used in the test anyway)

Addresses #387


libzeth/core/bits.cpp
```diff
@@ -62,4 +62,13 @@ std::vector<bool> bit_vector_from_size_t_be(size_t x)
     return res;
 }
 
+void bit_vector_write_string(const std::vector<bool> &bits, std::ostream &out_s)
```
AntoineRondelet, line 65: I'd rename to `bit_vector_write` for consistency with other serialization functions (e.g. https://github.com/clearmatics/libff/blob/develop/libff/algebra/fields/field_serialization.hpp#L33)

libzeth/tests/circuits/prfs_test.cpp
```diff
@@ -117,7 +119,7 @@ TEST(TestPRFs, TestPRFAddrApkGadget)
             0, 0, 0, 0, 0, 0, 0, 0, // 00
             1, 1, 1, 1, 1, 1, 1, 1  // FF
         },
-        ZERO);
+        "sk");
```
AntoineRondelet, line 122: Why not stick to the variable name here for the annotation? (same question for the lines below, e.g. why "expected0"/"expected1" and not "rho_expected0" etc.?)

libzeth/tests/circuits/sha256_test.cpp
```diff
@@ -180,26 +170,16 @@ TEST(TestSHA256, TestHashWithZeroLeg)
         pb, {left, right}, "Block_variable");
 
     libzeth::sha256_ethereum<Field> hasher(
-        pb, ZERO, input_block, result, "Sha256_ethereum");
+        pb, input_block, result, "Sha256_ethereum");
```
AntoineRondelet, line 173: `Sha256_ethereum` -> `sha256_ethereum` (I know that was like this before, but we don't use caps in annotations so one to fix here)

libzeth/circuits/circuit_utils.tcc
```diff
@@ -46,22 +46,21 @@ libsnark::linear_combination<FieldT> packed_addition(
...
         libsnark::pb_variable_array<FieldT>(inputs.rbegin(), inputs.rend()));
-};
+}
 
-// Takes a vector of boolean values, and convert this vector of boolean values
-// into a vector of FieldT::zero() and FieldT:one()
+// Allocate an array of variables on a given protoboard, and set the values to
+// zero or one based on a vector of bits.
 template<typename FieldT>
 libsnark::pb_variable_array<FieldT> variable_array_from_bit_vector(
```
AntoineRondelet, line 54: Actually, since the `fill_with_bits` method of `pb_variable_array` is used here, can't we simply remove the: `bits<numBits>::fill_variable_array` function? Looks like it is redundant since every call of `bits<numBits>::fill_variable_array` will be of the form: `<bit_vector>.fill_variable_array(pb, <pb_variable_array>);` and can thus be expressed by switching the pb_var and the bit vector as `<pb_variable_array>.fill_with_bits(pb, <bit_vector>);` 

Do you remember why the `bits<numBits>::fill_variable_array` function was needed (@dtebbs)?

Note: If still needed it would probably be more appropriate to rename `fill_variable_array` to `fill_pb_variable_array` to stick to type names, but the changes related to this function would need to be done in another PR anyway since it goes beyond the scope of this specific PR (`bits.fill_variable_array` is used in many gadgets)

dtebbs: For now, `bits<numBits>` is not compatible with `pb_variable_array<FieldT>::fill_with_bits`, so we would have to overload or modify this function in order to replace `bits<numBits>::fill_variable_array`.
Or, it could make sense to create an equivalent function to `variable_array_from_bit_vector` in this repo.

As you say, it would touch a lot of code, so maybe one for another PR.

libzeth/core/bits.cpp
dtebbs: Where the serialization functions use a specific encoding, the convention is `<type>_<operation>_<encoding>`.  In libff, we have functions which are parameterized by the encoding, hence we drop the `_<encoding>` part in the name.

libzeth/circuits/circuit_utils.tcc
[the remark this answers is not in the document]
dtebbs: Renamed everywhere.

libzeth/tests/circuits/prfs_test.cpp
dtebbs: changed the annotation string

libzeth/tests/circuits/sha256_test.cpp
dtebbs: changed the annotation

dtebbs:
Thanks for the comments @AntoineRondelet . Should be addressed in the latest version.

libzeth/core/bits.cpp
AntoineRondelet: Ah yes indeed, thanks

libzeth/tests/circuits/prfs_test.cpp
AntoineRondelet: The same comment applied for all the file where I can see that annotations differ from the variable names in many places.

libzeth/circuits/circuit_utils.tcc
AntoineRondelet: >> Do you remember why the bits<numBits>::fill_variable_array function was needed (@dtebbs)?
> For now, bits<numBits> is not compatible with pb_variable_array<FieldT>::fill_with_bits

Ah yes the types won't match..
From what I can see, `fill_with_bits`: https://github.com/clearmatics/libsnark/blob/master/libsnark/gadgetlib1/pb_variable.tcc#L49-L57 and `fill_variable_array`: https://github.com/clearmatics/zeth/blob/develop/libzeth/core/bits.tcc#L107-L119 are running the same code. Only diff is indeed on the type side of things, one is manipulating a vector, the other is manipulating an array. So `bits<numBits>::fill_variable_array` can be expressed as the composition of `array_to_vector` and `pb_variable_array<FieldT>::fill_with_bits` (i.e. its definition can be expressed as a wrapper around this function composition - introducing 2 additional function calls and context switching. This code is far from being the bottleneck though, so no major deal here). Anyway, not a big one, but I think we should rename the functions to strictly stick to the `<typeX>_to_<typeY>` syntax, and maybe consider some minor refactoring on the `bits` class code and/or `pb_variable` code. One for future PRs.

dtebbs: > So bits<numBits>::fill_variable_array can be expressed as the composition of array_to_vector and pb_variable_array<FieldT>::fill_with_bits (i.e. its definition can be expressed as a wrapper around this function composition - introducing 2 additional function calls and context switching. This code is far from being the bottleneck though, so no major deal here)

For the future but just to note, composing those functions would mean allocating and copying these structures, no matter how larger they are, just to deal with type differences.  A generic `fill_with_bits` function that can accept different containers would allow us to combine the code and avoid unnecessary allocating and copying.

AntoineRondelet: > A generic fill_with_bits function that can accept different containers would allow us to combine the code and avoid unnecessary allocating and copying.

Oh yes definitely, this would be much better

AntoineRondelet (approved):
LGTM, thanks @dtebbs

PR #392, opened 2021-08-17
CLIs name inconsistencies

AntoineRondelet:
- Add some missing flags to CLIs
- Stick to Kebab case for all CLIs for consistency (including for executable bash scripts)
- Add missing binaries to the PATH
- Add missing shebang
- Extend the CODING_STANDARDS to document the naming conventions for executables and structure in conditional tests.


libzeth/core/bits.hpp
```diff
@@ -35,7 +35,7 @@ template<size_t numBits> class bits : public std::array<bool, numBits>
 
     /// Construct from initializer-list.
     // cppcheck-suppress noExplicitConstructor
-    template<typename... boolList> bits(const boolList &...bits);
+    template<typename... boolList> bits(const boolList &... bits);
```
AntoineRondelet, line 38: Hum... odd, running `./scripts/format` locally seems to add this space which seems to break the format running on the CI. Need to have a look into this.

dtebbs: Worth checking the version of clang-format. (I'm using version 11).

CODING_STANDARDS.md
```diff
@@ -8,6 +8,20 @@ New code should follow these rules, and old code should be changed over time to
 - Each file should have a well-defined purpose and be concerned with a single narrow concept.
 - Add unit tests for new code. As far as possible, one test per source file or logical group of files.
 - Names should be as clear and concise as possible, avoiding opaque abbreviations.
+- In conditions, the tested variable must appear on the left of the operator. This is to stay consistent with English, which writes/reads from left to right.
+  ```
+  // GOOD:
+  if (age >= 18) { // Reads "If age is above 18" (the test is on the value of `age` here, so it appears first)
```
dtebbs, line 14: Could be worth moving / shortening / reformatting the comment to make it a bit more readable in a code box.

dtebbs (commented):
In order to make this uniform, we'll need to update all references in scripts etc, and we should also consider covering scripts themselves (the command name should not depend on the language used to author it).

However, this conflicts with https://github.com/clearmatics/neodizk/pull/36#discussion_r691299416.

AntoineRondelet:
> In order to make this uniform, we'll need to update all references in scripts etc, and we should also consider covering scripts themselves (the command name should not depend on the language used to author it).
> 
> However, this conflicts with [clearmatics/neodizk#36 (comment)](https://github.com/clearmatics/neodizk/pull/36#discussion_r691299416).

Good point I didn't take the scripts into consideration here, was focusing on our C++ and Python CLIs. So we have even more inconsistencies then.. All our file names are `snake_case` (bash, python, c++) so all bash scripts have a `_` as separator. However, all commands from the CLIs (except `prover_server`) are in `kebab-case`, so it looks like the way to go is to use `snake_case` everywhere then.

CODING_STANDARDS.md
AntoineRondelet: Restructured to a multiple line comment

AntoineRondelet:
I get an infinite loop on the CI because `server_start` keeps waiting for the `prover-server` to be active. Have a look closer to the CI scripts, I must have forgotten some renamings here.

dtebbs:
> I get an infinite loop on the CI because `server_start` keeps waiting for the `prover-server` to be active. Have a look closer to the CI scripts, I must have forgotten some renamings here.

Looks like some missing renamings in the ci_utils.sh file  (e.g. line 111).

dtebbs (approved):
LGTM

AntoineRondelet:
> > I get an infinite loop on the CI because `server_start` keeps waiting for the `prover-server` to be active. Have a look closer to the CI scripts, I must have forgotten some renamings here.
> 
> Looks like some missing renamings in the ci_utils.sh file (e.g. line 111).
> 
> UPDATE: fixed in the latest commit

Thanks @dtebbs

PR #404, opened 2021-09-21
Util script fix

dtebbs:
Issue in ci util scripts (ci_utils.sh executes instead of sourcing build_utils.sh, which should not work).  This issue was shown up when trying to update zeth in zecale.  It's not clear to me why this works in this repo, but not in zecale.


AntoineRondelet (approved):
LGTM, thanks @dtebbs

PR #416, opened 2021-11-03
Updated libsnark submodule to deprecate git:// links

AntoineRondelet:
As per the title, follows from: https://github.com/clearmatics/libsnark/pull/37


dtebbs (approved):
LGTM

Issue #417, opened 2021-11-04
Better expose prover_server (generic) util functions


AntoineRondelet:
Related to https://github.com/clearmatics/zecale/pull/89#discussion_r742720040

Despite being mere wrappers around the libzeth functions, some functions in the `prover_server` (Zeth) and `aggregator_server` (Zecale) are duplicated across the projects (functions like `load_keypair`, `write_keypair`, `write_constraint_system` etc). Duplicated code is prone to inconsistencies (one version changed and not the others) which is not good. Also, with code: less is more. Less code is less code to maintain and always a good idea. Let's see how we could package these utils function on the prover server to better expose them and use them straight in other projects like Zecale. Maybe we could create a new namespace like `libzeth::server` that contains all these functions and we could use them straight in `prover_server.cpp` in Zeth and in `aggregator_server.cpp` in Zecale.

dtebbs:
I was just addressing this as part of https://github.com/clearmatics/zecale/pull/89#discussion_r742720040. The main reason for these functions was to open the streams with the error-checking options (which is a couple of pretty annoying lines).  This has now been wraped in the libtool library, so these functions are no longer required (and removing them results in less code).

See #418

PR #418, opened 2021-11-04
Clean up serialization and stream handling in prover_server

dtebbs:
Remove some tiny wrapper functions which are no longer necessary (and duplicated in other projects - thanks @AntoineRondelet for pointing this out).

Their main function was to wrap the creation of fiile streams with error-checking options enabled, which can now be done in a single line via `libtool::open_binary_*_file` functions.

### DIFF
--- a/prover_server/prover_server.cpp
+++ b/prover_server/prover_server.cpp
@@ -2,6 +2,7 @@
 //
 // SPDX-License-Identifier: LGPL-3.0+
 
+#include "libtool/tool_util.hpp"
 #include "libzeth/circuits/circuit_types.hpp"
 #include "libzeth/core/extended_proof.hpp"
 #include "libzeth/core/utils.hpp"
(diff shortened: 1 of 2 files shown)


dtebbs:
Addresses #417

prover_server/prover_server.cpp
```diff
@@ -520,34 +455,49 @@ int main(int argc, char **argv)
...
             std::cout << "[DEBUG] Writing separate verification key to "
                       << verification_key_output_file << "\n";
-            write_verification_key(keypair.vk, verification_key_output_file);
+            std::ofstream out_s = libtool::open_binary_output_file(
+                verification_key_output_file.c_str());
+            snark::verification_key_write_bytes(keypair.vk, out_s);
         }
 
+        // Write the keypair last. If something above fails, this same
+        // code-path will be executed again on the next invocation.
```
AntoineRondelet, line 486: Just for my understanding what's the worry here if the keypair is written first?
Do you fear that in the event of an error a keypair could fail to be written on file entirely but the corresponding pk could have already been written leading to some potential inconsistencies in what's written to files?

AntoineRondelet (approved):
Great thanks for this @dtebbs - good to see some code go away!
LGTM overall, please just see my question below.

prover_server/prover_server.cpp
dtebbs: If the keypair is written first, but something fails afterwards, we would have to manually remove the keypair and re-run in order to re-try the other operations (since the control flow depends on the existence of the keypair).

PR #423, opened 2022-04-01
Ci fix

dtebbs:
Fix some sudden CI breakages:
- [x] grpc repo uses `git://` references for submodules (no longer supported by public github access)
- [x] flask has loose dependency specifications, and pulls in new incompatible versions of several dependencies.  (Note, I tried upgrading to a later minor version, but this resulted in similar dependency issues.  For simplicity I've added package restrictions in the mpc `setup.py`, whcih can be removed if we upgrade the flask dependency in the future.)

### DIFF
--- a/mpc/setup.py
+++ b/mpc/setup.py
@@ -33,6 +33,11 @@ setup(
         "ecdsa==0.13.3",
         "click==7.0",
         "requests==2.22.0",
+        # flask constraints on dependencies are too loose. Constrain the
+        # versions here to avoid CI failures.
+        "Jinja2==3.0.3",
+        "itsdangerous==2.0.1",
+        "Werkzeug==2.0.2",
     ],
     scripts=[
         "commands/phase1_server",
--- a/.github/workflows/dev-docs.yml
+++ b/.github/workflows/dev-docs.yml
@@ -12,10 +12,10 @@ jobs:
     - name: Cache grpc
       uses: actions/cache@v2
       with:
-        key: grpc-1.31.x-${{ runner.os }}
+        key: grpc-1.32.x-${{ runner.os }}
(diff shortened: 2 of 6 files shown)


mpc/setup.py
```diff
@@ -33,6 +33,11 @@
         "ecdsa==0.13.3",
         "click==7.0",
         "requests==2.22.0",
+        # flask constraints on dependencies are to loose. Constrain the version
```
AntoineRondelet, line 36: - `are to loose` -> `are too loose`?
- `Constrain the version them here` -> I think this needs rephrasing, maybe `constrain the versions here to avoid [...]`?

AntoineRondelet:
@dtebbs out of curiosity, why didn't you bump gRPC further? The newest release is 1.45.1 (https://github.com/grpc/grpc/releases). Did you see any conflict with the newest releases? If not, we probably want to bump to a more recent release to bring some optimizations and security fixes in.

mpc/setup.py
dtebbs: Ah yes.  Thanks for that.
Updated.

dtebbs:
> @dtebbs out of curiosity, why didn't you bump gRPC further? The newest release is 1.45.1 (https://github.com/grpc/grpc/releases). Did you see any conflict with the newest releases? If not, we probably want to bump to a more recent release to bring some optimizations and security fixes in.

Yes, it does make sense to upgrade at some point.  For now I didn't want to get into potential API changes etc requiring chages in our code, so took the most conservative approach.  But at some point we should upgrade.  I've made a branch `https://github.com/clearmatics/zeth/tree/grpc-upgrade` to run the CI on a much newer version.

